PR #1, opened 2015-10-18
Really requires PLY.

sfermigier:
This forces installation of the PLY package when doing a `pip install`.

Also, what's constraint on the Python version ? This should probably be added to the `setup.py` too.


gvanrossum:
@matthiaskramm Can you manually merge this if something still needs to be done and/or just close it?

matthiaskramm:
Merged manually. @sfermigier, could you check whether this does what you expected?

(It doesn't help with https://github.com/google/pytype/issues/6, which is a different issue)

PR #20, opened 2016-01-30
Make pytype installable via pip

### LINKED ISSUE #6: python -m pip install -U . doesn't work.
It ought to be possible to install pytype using pip by running

```
python -m pip install -U .
```

but doing so causes an error message.

berkerpeksag:
An alternative solution would be to define a `recursive-include` rule in `MANIFEST.in`.

Note: `setup.py` uses `install_requires`, but it's a setuptools feature and setuptools is not imported there.

Closes #6


matthiaskramm:
Merging, thanks for the patch!

PR #24, opened 2016-06-01
Document some types in __builtins__.

durin42:


martindemello:
these have been merged manually as part of other changes; could you sync to head and check?

PR #27, opened 2016-07-15
add pytype.pytd.builtins to setup.py

tharvik:
pytest need to access `pytype/pytd/builtins/__builtin__.py` at runtime, but it isn't installed explicitly, this PR fixes that.


matthiaskramm:
Closing since we don't have a CLA.

PR #48, opened 2017-02-15
Support loading of 3rd party stubs from typeshed.

gvanrossum:
Also fix an inconsistency in how files are actually loaded.

This is most likely wrong; I'd be happy to see how this should be done. There are really two separate fixes here:

- The fix to load_pytd.py tries to load a module from the typeshed/third_party tree before trying stdlib. (It also adds an XXX comment pointing out that the logic for finding builtin modules doesn't work any more for typeshed. But perhaps that doesn't matter, because there are only a few modules that require this behavior, and presumably you have pytd files for those still.)
- The other fix makes the way `Typeshed._load_file()` loads a file consistent: the original code was hardcoding the pathname but then returning the pathname computed from `self._typeshed_path`. But maybe that was intentional? This didn't actually break me AFAIK, but I ran into the inconsistency when tracing through the code.

Question: How are tests run? There's nothing in your README.md about that.

### DIFF
--- a/pytype/load_pytd.py
+++ b/pytype/load_pytd.py
@@ -243,6 +243,7 @@ class Loader(object):
     log.debug("Trying to import %r", module_name)
     # Builtin modules (but not standard library modules!) take precedence
     # over modules in PYTHONPATH.
+    # XXX typeshed no longer has a builtins subdir.
     mod = self._load_builtin("builtins", module_name)
     if mod:
       return mod
@@ -250,6 +251,11 @@ class Loader(object):
     file_ast = self._import_file(module_name, module_name.split("."))
     if file_ast:
       return file_ast
+
+    # Try a third party module (typically site-packages).
+    mod = self._load_builtin("third_party", module_name)
+    if mod:
+      return mod
 
     # The standard library is (typically) at the end of PYTHONPATH.
     mod = self._load_builtin("stdlib", module_name)
--- a/pytype/load_pytd.py
+++ b/pytype/load_pytd.py
@@ -243,7 +243,8 @@ class Loader(object):
     log.debug("Trying to import %r", module_name)
     # Builtin modules (but not standard library modules!) take precedence
     # over modules in PYTHONPATH.
-    # XXX typeshed no longer has a builtins subdir.
+    # Note: while typeshed no longer has a builtins subdir, the pytd
+    # tree still does, and order is important here.
     mod = self._load_builtin("builtins", module_name)
     if mod:
       return mod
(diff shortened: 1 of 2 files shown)


matthiaskramm:
Regarding the `XXX`: We do still have a few builtins, in `pytd/builtins/*.pytd`. And order actually matters for them (they need to have higher priority than "user space" pyi). Mostly because of `typing.pytd`.

Also, in order for this to not break our internal use of pytype, it should probably be controlled by a flag, or, at least, it shouldn't happen if a self.options.imports_map is not None. I'll add that.

Also, I'm curious: why `third_party` before `stdlib`? Not saying another ordering is better, but do you have a use case in mind that would break if we try `stdlib` before `third_party`?

Other than that, this change looks good. Happy to merge it into pytype. (Of course, I can't actually do that in github, since our open-source export is unidirectional - but I can convert it to a change against our internal code review system, get it code reviewed, merge it in, and then export a new version of pytype that has it. You might want to use your patched version in the mean time.)

gvanrossum:
FWIW, here's how I tested this. Consider the file `j.py`:
```python
from simplejson import dumps, loads

def dork(x):
    return dumps(x)

def krod(s):
    return loads(s)
```
Then check it as
```
mypy -o- j.py
```
Before this PR the output is:
```
from typing import Any

dumps = ...  # type: Any
loads = ...  # type: Any

def dork(x) -> Any: ...
def krod(s) -> Any: ...
File "j.py", line 1, in <module>: Can't find module 'simplejson'. [import-error]
```
(The last line is an error written to stderr; what comes before is the generated stub.)

With this PR, there is no error, and the stub output is:
```python
from typing import Any, Dict

def dork(x) -> str: ...
def dumps(obj, *args, **kwds) -> str: ...
def krod(s: str) -> Any: ...
def loads(s: str, **kwds) -> Any: ...
```
So it has found the `simplejson` stub (which lives in `typeshed/third_party/2/simplejson/`).

gvanrossum:
Our messages crossed. Feel free to change the XXX comment to something closer to the truth (I did qualify it with `typeshed`).

Re: putting third_party before stdlib, I got this backwards, sorry. Should I upload a fixed commit?

matthiaskramm:
Thanks for doing the manual test! I'll add some actual test cases based on what you wrote.

Feel free to leave things as-is, I'll switch the order of third_party and stdlib when I apply this.

As for running the tests, we probably need to supply some `run_tests.py` script (we use Bazel to run them internally, but let's not make open-source users install that). I suspect you'd need a patched Python 2 interpreter for actually running *all* of them, though (e.g. tests/test_annotations.py).

gvanrossum:
Cool. The manual test was actually the essence of the problem that led me
here... FWIW for Dropbox running bazel isn't a problem.

matthiaskramm:
Applied this manually: https://github.com/google/pytype/commit/8f2f82b6b4a6074e69d44e17dd2c63aa2c0b3aba

PR #85, opened 2018-05-18
Update typeshed submodule.

sivachandra:


sivachandra:
Huh! I pushed without updating my email address from my laptop and I think I kind of stuck with this. Lets close this and I will create another pull request.

PR #91, opened 2018-06-21
Report import errors due to bad deps.

rchen152:
PiperOrigin-RevId: 201481602


rchen152:
Hey Siva, continuous-integration/travis-ci/pr failed with an error about cmake installation. Since this PR does include a change to test_module, I'm not sure if the failure is relevant or not. Could you take a look?

sivachandra:
The failure is akin to a TAP tool failure. So, you should just click the "Restart Job" on the top right. I have done that now and it looks like order has been restored :)

sivachandra:
If you did not already notice, the branch was green but the pull request went red.

rchen152:
Ah, got it. Thanks!

PR #93, opened 2018-06-29
Project import generated by Copybara.

rchen152:
Unfortunately, I had to run the import in SQUASH mode because of two problems:
* buildutils had been renamed to build_scripts, so we can't use the copybara config at HEAD. Usually, passing --read-config-from-change to use the config at every change to import that change would fix this, but:
* at several points, our config is broken. There was a leakr problem, which is solved by using --leakr-dictionary to manually pass in the correct leakr.dic from HEAD, but I had also broken the config by not removing `__future__.pytd` everywhere it needed to be deleted, and I couldn't figure out how to get around this one.

I believe this is a one-time problem that we won't have again, since the Copybara presubmit should prevent us from submitting broken states in the future.


sivachandra:
LGTM, but I will wait for the checks to turn green before approving.

[You could merge yourselves without any approval, but I think we should set examples.]

I have a question: When I renamed the directory, I did all my testing with copybara. So, can you explain a bit more about the problem you are seeing because of the rename.

rchen152:
The copybara config at HEAD was being used to attempt to migrate the changes from before the rename, so there were errors about build_scripts not existing.

PR #96, opened 2018-07-17
Install python3.5-dev and python3.6-dev.

sivachandra:
This is required so that Travis' virutalenvs can setup the include directory
correctly. Pytype's extension modules can then be built pointing to the
header files corresponding to the correct Python version.


sivachandra:
This change should unblock Rebecca's PR.

PR #98, opened 2018-07-17
Google sync

rchen152:
Using the copybara config at HEAD didn't work because we've changed the file too many times recently as we got rid of all our stripping.

Instead, I pushed all the commits up to and including yours that made copybara stop tracking .travis.yml and build_scripts/. For whatever reason, that commit added the deleted lines back to .travis.yml (yay) but deleted build_scripts/ entirely (boo). So I pushed another commit adding that directory back. Since we won't be making any internal build_script/ changes going forward, I think this was safe to do.


sivachandra:
Re-running is not helping here. I am digging again.
To restart a job, click on the "Restart job" button on the top right half on the job page, https://travis-ci.org/google/pytype/jobs/405142678

sivachandra:
Yes, it is the segfault problem again. I forked pytype and ran a build with detailed logs. This is what I got: https://travis-ci.org/sivachandra/pytype/jobs/405168004. Notice the segfault at the end of the log.

My refcounts change is speculative; not sure it will fix the bug, but hope it does. About this PR, I am approving it. Up to you if you want to submit or not. Especially, if you are going to follow up with submitting the refcount change, I would say go ahead and submit.

rchen152:
Thanks for confirming; I'll go ahead and submit. (Also, I was quite confused because I didn't see the "Restart Job" button you referred to, but it turned out that was because I'd never signed into travis-ci.org!)

PR #99, opened 2018-07-18
Google sync

rchen152:
Unfortunately, the segfault bug appears to still be present.


sivachandra:
Sorry, that it is not fixed yet. I will keep looking. But, put this in anyway. I am also working on improving the diagnostic messages when a test run fails.

PR #100, opened 2018-07-18
Log cmake and ninja outputs, and add more diagnostic messages to stdout.

sivachandra:


build_scripts/build_utils.py
```diff
@@ -11,6 +11,8 @@
 OUT_DIR = os.path.join(PYTYPE_SRC_ROOT, "out")
 SRC_PYI_DIR = os.path.join(PYTYPE_SRC_ROOT, "pytype", "pyi")
 OUT_PYI_DIR = os.path.join(OUT_DIR, "pytype", "pyi")
+CMAKE_LOG = os.path.join(OUT_DIR, "cmake.log")
```
rchen152: I've been wondering, is there a way to see the logs from a Travis run? I see /home/travis/... paths, but there's no obvious way to access them.

build_scripts/run_tests.py
```diff
@@ -52,62 +56,81 @@ def print_report(self):
...
-    if mod_name:
-      print(l)
-    if log_file:
-      have_failures = True
-      if fail_collector:
-        fail_collector.add_failure(mod_name, log_file)
+  failed_targets = []
+  with open(build_utils.NINJA_LOG, "w") as ninja_log:
+    while True:
+      l = process.stdout.readline()
```
rchen152: nit: lowercase L looks like number 1; maybe call this `line` or something?

sivachandra: Done.

build_scripts/build_utils.py
sivachandra: As discussed offline, we should setup uploading build artifacts to GCS so that we can download and see the logs.

PR #101, opened 2018-07-20
Google sync

rchen152:
Pulling in Siva's reference leak fixes to see if they resolve our test issues.


martindemello:
still getting travis failures, sadly.

sivachandra:
I have at least three more changes before we can understand what exactly is going on :)

PR #109, opened 2018-08-02
Add a convenience script to build all pytype targets.

sivachandra:
This script will be expanded in a later pass to add options to build
the executables like 'pytype' and 'pytype-single' with C++ logging
enabled.

### DIFF
--- a/build_scripts/build_utils.py
+++ b/build_scripts/build_utils.py
@@ -22,6 +22,11 @@ GEN_FILE_LIST = [
     "stack.hh",
 ]
 
+NINJA_FAILURE_PREFIX = "FAILED: "
+FAILURE_MSG_PREFIX = ">>> FAIL"
+PASS_MSG_PREFIX = ">>> PASS"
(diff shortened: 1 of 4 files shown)


build_scripts/build_utils.py
```diff
@@ -58,6 +63,27 @@ def _clean_out_dir(msg):
       os.remove(path)
 
 
+def get_module_and_log_file_from_result_msg(msg):
+  if msg.startswith(FAILURE_MSG_PREFIX):
+    _, mod_name, log_file = msg.split(RESULT_MSG_SEP)
```
rchen152: Here you assume that the msg always has three components, but in failure_msg, adding the log file is optional.

sivachandra: Good catch and thanks for it. It showed me that a mere copy paste from the other file is not good enough. I restructured the code a bit to make it cleaner and fix the problem you have pointed out.

PR #113, opened 2018-08-03
Google sync

Solumin:
Resolves #112 
(It looks like the message on 0e440c2 won't be enough to close the PR, because it's missing the `#`. So I'm editing the description to close it instead. This bug has been fixed upstream.)

### DIFF
--- a/cmake/modules/PyTypeUtils.cmake
+++ b/cmake/modules/PyTypeUtils.cmake
@@ -41,6 +41,11 @@ add_custom_target(
   ${ALL_TESTS_TARGET}
 )
 
+string(COMPARE EQUAL ${CMAKE_BUILD_TYPE} "Debug" is_debug_build)
+if(is_debug_build)
+  add_definitions(-DPYTYPE_ENABLE_CPP_LOGGING)
+endif(is_debug_build)
+
 # Returns the fully qualified parent name for targets in the current source
 # directory in |fq_parent_name|.
 function(_gen_fq_target_parent_name fq_parent_name)
--- a/pytype/abstract.py
+++ b/pytype/abstract.py
@@ -933,7 +933,7 @@ class Instance(SimpleAbstractValue):
 
   def compatible_with(self, logical_value):  # pylint: disable=unused-argument
     # Containers with unset parameters and NoneType instances cannot match True.
-    name = self.get_full_name()
+    name = self.full_name
     if logical_value and name in Instance._CONTAINER_NAMES:
       return (T in self.type_parameters and
               bool(self.type_parameters[T].bindings))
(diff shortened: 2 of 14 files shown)


rchen152:
Looks like there's a real test failure this time:
```
CMake Error at cmake/modules/PyTypeUtils.cmake:44 (string):
  string sub-command COMPARE, mode EQUAL needs at least 5 arguments total to
  command.
Call Stack (most recent call first):
  CMakeLists.txt:10 (include)
```
https://travis-ci.org/google/pytype/jobs/411937114

Solumin:
That looks like Siva's C++ logging commit.

cmake/modules/PyTypeUtils.cmake
```diff
@@ -41,6 +41,11 @@ add_custom_target(
   ${ALL_TESTS_TARGET}
 )
 
+string(COMPARE EQUAL ${CMAKE_BUILD_TYPE} "Debug" is_debug_build)
```
Solumin: Could this be replaced with `if(${CMAKE_BUILD_TYPE} STREQUAL "Debug")`?

sivachandra: I have sent an internal CL to fix this. However, can you add a commit to this PR and let copybara import it? The fix you suggested should also work.

sivachandra: If you are going to attempt this, you will need quotes like this: "${CMAKE_BUILD_TYPE}". I preferred "string(COMPARE ...)" over the "if(...)" because CMake is what it is: https://cmake.org/pipermail/cmake/2013-February/053587.html. The suggested alternate there also has problems. The issue is that "if" is not a statement but a function which needs to be matched by another function "endif"!

CMake also has an expectation that we write it this way:

if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
...
endif("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")

So, as a practice, I have been using "string(COMPARE ...)" followed by an "if(...)".

sivachandra: Just noticed that you have actually assigned this to me. I am not sure what that means, but as I had asked earlier, can I push a commit to this branch (and expect copybara to do the right import?)

Solumin:
I assigned it to you because I wanted to make sure you saw the failure,
since you're the most familiar with CMake.

I'm fairly certain Copybara will not do the right thing, since this is an
internal-to-GitHub push branch, not a pull request. I think the best course
of action is discarding this PR, doing an internal change and then
reexporting. (You could also do separate PR instead of an internal change,
I'll just have to import that PR first.)

sivachandra:
I have an internal change sent out to fix the problem. I can do a new copybara export PR after that lands.

I apologize for improper testing here. I tested that it works when a build type is specified, but did not test that it breaks when a build type is not specified. Sine the original change was in google3, I ran out of patience with chromoting and just submitted it :(

Solumin:
Internal change is ready to land.

No worries, mistakes happen and this is a very minor one. (the tests are
already failing anyway...)

sivachandra:
Can I close this PR, delete the branch, and create another copybara export PR?

Solumin:
Sounds good to me.

PR #114, opened 2018-08-06
Add --debug option to the scripts build.py and run_tests.py

sivachandra:
Specifying this option builds all Pytype targets in the debug mode. It is useful, for example, to enable C++ logging.


sivachandra:
The real effect of this change can be seen after this PR lands: https://github.com/google/pytype/pull/113

PR #117, opened 2018-08-09
Allow pytype to override an existing logging level.

rchen152:
When pytype is called from analyze_project, we need to be able to
override analyze_project's own logging level in order to use the one
requrested for pytype.

This will have to go through the github->internal->github workflow, but I figured that it would be nice to make open-source-related changes on GitHub.

### DIFF
--- a/pytype/config.py
+++ b/pytype/config.py
@@ -408,7 +408,11 @@ class Postprocessor(object):
       # "verbosity=-1" can be used to disable all logging, so configure
       # logging accordingly.
       basic_logging_level = logging.CRITICAL + 1
-    logging.basicConfig(level=basic_logging_level)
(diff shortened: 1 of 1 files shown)


Solumin:
I don't understand the motivation for this change. Why do analyze_project and pytype have different logging levels in the first place?

rchen152:
Because we often want to use a less noisy logging level for pytype than for analyze_project. For example, when I request logging.INFO messages for analyze_project, I want to see things like the configuration options from setup.cfg, not the entire bytecode execution flow for every file.

Solumin:
Ah, I see! That makes sense.

I believe GitHub is a bit behind HEAD right now, so I'd like to sync before submitting this.

PR #118, opened 2018-08-09
Google sync

Solumin:


rchen152:
Ha, looks like we have a real failure this time - test_errors is failing: https://travis-ci.org/google/pytype/jobs/414156834

rchen152:
Looks like it's because we added a test case with type annotations; I'll fix it.

Solumin:
Discarding this PR then. We'll re-push once that's fixed.

PR #120, opened 2018-08-09
Fix compiler warnings due to assigning const char* values to char* variables.

sivachandra:
This change essentially makes the variables also of type const char*,
but uses const_cast at all relevant places where these variables are used
as char* values.


sivachandra:
Since we "see" the compiler warnings when building via CMake, I prepared a PR instead of an internal change.

Solumin:
I'll pull this in when #122 is merged.

PR #124, opened 2018-08-10
Support excluding files or directories from analyze_project.

rchen152:
* Supports an --exclude flag to make it easy to exclude parts of
  a directory that are not pytype-safe (e.g.,
  `pytype subpackage --exclude subpackage/*_test.py`)
* Renames the filenames arg to inputs in preparation for adding
  it to the configuration file, since `filenames` is a bit of a
  misnomer when we allow directories as well.

This PR will have to be imported into Google and then re-exported.

### DIFF
--- a/pytype/tools/analyze_project/main.py
+++ b/pytype/tools/analyze_project/main.py
@@ -50,7 +50,9 @@ def main():
     config.generate_sample_config_or_die(args.generate_config)
     sys.exit(0)
 
-  args.filenames = file_utils.expand_source_files(args.filenames)
+  # Empty posargs defaults to a list, but an empty nargs flag defaults to None.
+  args.inputs = (file_utils.expand_source_files(args.inputs) -
+                 file_utils.expand_source_files(args.exclude or []))
   conf = parser.config_from_defaults()
   # File options overwrite defaults.
   file_config = config.read_config_file_or_die(args.config)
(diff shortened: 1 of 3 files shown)


pytype/tools/analyze_project/main.py
```diff
@@ -59,10 +61,10 @@ def main():
   # Command line arguments overwrite file options.
   conf.populate_from(args)
   if not conf.pythonpath:
-    conf.pythonpath = environment.compute_pythonpath(args.filenames)
+    conf.pythonpath = environment.compute_pythonpath(args.inputs)
   logging.info('\n  '.join(['Configuration:'] + str(conf).split('\n')))
 
-  if not args.filenames:
+  if not args.inputs:
     parser.parser.print_usage()
```
martindemello: while you're at it, add a print statement here saying "No input files" - just printing out the usage can be unhelpful, especially if we add inputs to the config file.

rchen152: Done.

rchen152:
This PR changed files that are tracked internally, so it should've gone through the Copybara PR workflow rather than being merged directly. (I noted that in the initial description.) I *think* to fix things now I can just make the same change internally and Copybara will figure everything out; let me try.

martindemello:
ah, i missed that comment in the initial description, sorry.

PR #130, opened 2018-08-15
Add googletest as a submodule.

sivachandra:


sivachandra:
An internal change will add cc_test targets making use of this new submodule.

PR #136, opened 2018-08-20
Add a simple script to run a C++ unit test binary.

sivachandra:
This script will be called from the CMake cc_test rule.

[Changes to the rule to make use of this script will be made as a
Google internal change and exported to Github.]


sivachandra:
Ping. This block the piper change 209578650.

PR #139, opened 2018-08-29
Google sync

sivachandra:


Solumin:
The load_pyi change hasn't been released internally yet, so I'm a little uncertain about syncing it here.

sivachandra:
This is not a release; its only a commit :)
If the tests are clean, I do not see any reason not to submit.

Solumin:
There's a been a number of times where our tests were clean but a change caused a crash.

But yes, this isn't a release, unless someone is pulling from HEAD.

PR #141, opened 2018-09-04
Add cpython submodule pinned to the current 2.7 HEAD.

sivachandra:
Extended the build script to build the patched Python2.7 interpreter if
the option "--py27" is specified. A standalone script which can build
the patched interpreter has also been added. This script will be used by
a CMake target to build the patched interpreter as part of the
run_tests.py flow.

### DIFF
--- a/build_scripts/build.py
+++ b/build_scripts/build.py
@@ -10,8 +10,10 @@ Specifying the -c option forces a clobber before building.
 from __future__ import print_function
 
 import argparse
+import sys
 
 import build_utils
+import py27
 
 def parse_args():
   parser = argparse.ArgumentParser()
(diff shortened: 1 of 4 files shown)


build_scripts/build.py
```diff
@@ -10,27 +10,34 @@
...
 
 def parse_args():
   parser = argparse.ArgumentParser()
   parser.add_argument("--clobber", "-c", action="store_true", default=False,
                       help="Force clobber before building.")
   parser.add_argument("--debug", "-d", action="store_true", default=False,
                       help="Build targets in the debug mode.")
+  parser.add_argument(
+      "--py27", action="store_true", default=False,
+      help="Build Python-2.7 interpreter with type annotaion patch applied.")
```
rchen152: s/annotaion/annotation/

PR #142, opened 2018-09-04
Future-proof preconditions.py:_TOKEN_RE

### LINKED ISSUE #140: FutureWarning under Py3.7: Possible nested set (preconditions.py:166)
Running pytype under Python 3.7, there is a new warning about a regular expression:

```
/venv/lib/python3.7/site-packages/pytype/pytd/parse/preconditions.py:166: FutureWarning: Possible nested set at position 19
  r"\s*(?:(?P<literal>[[\]{}])|(?P<word>[a-zA-Z_]\w*))")
```

I think the fix is to escape the [ just inside the first character class:

```python
  r"\s*(?:(?P<literal>[\[\]{}])|(?P<word>[a-zA-Z_]\w*))")
```

Solumin:
3.7 adds a warning about possible future changes to re: https://bugs.python.org/issue30349
A future version of python will add nested sets, which allows nesting of sets using `[...]`. Escape the inner `[` in the re so it doesn't trigger a nested set.
Closes #140.

### DIFF
--- a/pytype/pytd/parse/preconditions.py
+++ b/pytype/pytd/parse/preconditions.py
@@ -163,7 +163,7 @@ class CallChecker(object):
 
 # RE to match a single token.  Leading whitepace is ignored.
 _TOKEN_RE = re.compile(
-    r"\s*(?:(?P<literal>[[\]{}])|(?P<word>[a-zA-Z_]\w*))")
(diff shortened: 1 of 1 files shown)


Solumin:
Would like to resolve #143 before importing this one.

Issue #134, opened 2018-08-18
pytype chokes on Python 3 yaml (upstream typeshed issue?)


lantz:
`pytype` seems to fail on the `yaml` library if it's used in Python 3 code:
```
$ cat yamltest.py
import yaml
$ pytype -V3.5 yamltest.py
Analyzing 1 sources with 0 dependencies
File ".../pytype-bugs/yamltest.py", line 1, in <module>: Couldn't import pyi for 'yaml' [pyi-error]
    File: "/usr/local/lib/python3.5/dist-packages/pytype/typeshed/third_party/2and3/yaml/__init__.pyi", line 39
      b'class YAMLObject:'
        ^
  ParseError: Illegal value for alias '__metaclass__'. Value must be an attribute or method on the same class.

For more details, see https://github.com/google/pytype/blob/master/docs/errors.md#pyi-error.
```

The root problem seems to be with `typeshed`: python/typeshed#2399

lantz:
Until it is fixed upstream this seems to work:
```python
import yaml  # pytype: ignore=pyi-error
```

rchen152:
The __metaclass__ issue has been fixed, but pytype's still unhappy with the nested `classobj` definition in yaml.constructor. I'll submit a PR to try to change that to `classobj: type`; if that doesn't work, we'll have to update our parser to handle nested classes.

rchen152:
Fixed and released!

PR #147, opened 2018-09-07
Add partial support for nested classes in pyi files.

rchen152:
This PR will have to be imported and re-exported rather than merged directly.

It attempts to resolve https://github.com/google/pytype/issues/134 by fixing the second issue that prevented pytype from parsing the yaml pyi files, the existence of a nested class in yaml.constructor. I decided to change our parser rather than typeshed because I found more examples of nested classes in typeshed/third_party/2and3/google/protobuf/.

I'm not yet calling the issue fixed because I still want to do a release (hence the `__version__` change), verify the fix, and then change typeshed's pytype_test to cover yaml/ going forward.

### DIFF
--- a/pytype/pyi/parser.py
+++ b/pytype/pyi/parser.py
@@ -251,12 +251,12 @@ class _Parser(object):
     new_constant()
     add_alias_or_constant()
     add_import()
+    new_class()
     new_type()
     new_union_type()
     new_function()
     new_named_tuple()
     regiser_class_name()
-    add_class()
     add_type_var()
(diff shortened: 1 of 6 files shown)


pytype/pyi/parser.py
```diff
@@ -946,6 +945,9 @@ def add_class(self, class_name, parent_args, defs):
       defs: A list of constant (pytd.Constant) and function (_NameAndSig)
```
Solumin: Can also contain classes now?

PR #156, opened 2018-09-13
Organize Travis commands into steps.

sivachandra:
Three steps have been added to build, test and type check pytype sources.

### DIFF
--- a/build_scripts/travis_script.py
+++ b/build_scripts/travis_script.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+"""A simple script to run Travis build steps."""
+
+from __future__ import print_function
+
+import collections
+import os
+import sys
+
+import build_utils
+
+def _ismod(f):
+  """Return True if f is a Python module but not a test module."""
+  return f.endswith(".py") and not f.endswith("_test.py")
+
+
+STEP = collections.namedtuple("STEP", ["name", "command"])
+
+PYTYPE_COMMAND = os.path.join("out", "bin", "pytype")
+PYC_DIR = os.path.join(build_utils.PYTYPE_SRC_ROOT, "pytype", "pyc")
+PYC_FILES = [os.path.join(PYC_DIR, f) for f in os.listdir(PYC_DIR) if _ismod(f)]
+TYPECHECK_FILES = [
+    "pytype/compat.py", "pytype/debug.py", "pytype/utils.py"] + PYC_FILES
+
+
+def _begin_step(s):
+  print("")
+  print("BEGIN_STEP: %s" % s.name)
+  print("STEP_COMMAND: %s" % ' '.join(s.command))
+  print("")
+
+
+def _end_step(s):
+  print("\nEND_STEP: %s\n" % s.name)
+
+
+def _report_failure(s):
+  print("")
+  print(">>> STEP_FAILED: %s" % s.name)
+  print("")
+
+
+def _run_steps(steps):
+  for s in steps:
+    _begin_step(s)
+    returncode, _ = build_utils.run_cmd(s.command, pipe=False)
+    if returncode != 0:
+      _report_failure(s)
+      sys.exit(1)
+    _end_step(s)
+    
+
+def main():
+  s1 = STEP(name="Build",
+            command=["python", build_utils.build_script("build.py")])
+  s2 = STEP(name="Run Tests",
+            command=["python", build_utils.build_script("run_tests.py"), "-f"])
+  s3 = STEP(name="Type Check",
+            command=[PYTYPE_COMMAND] + TYPECHECK_FILES)
+  _run_steps([s1, s2, s3])
+  print("\n*** All build steps completed successfully! ***\n")
+
+
+if __name__ == "__main__":
+  main()
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,21 +37,11 @@ addons:
     - python3.7-dev  # To build host extension modules under 3.7
 
 install:
+  - pip install importlab
   - pip install pyyaml
   - pip install six
-  - pip install pytype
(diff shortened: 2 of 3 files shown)


build_scripts/travis_script.py
```diff
@@ -0,0 +1,64 @@
...
+      sys.exit(1)
+    _end_step(s)
+    
+
+def main():
+  s1 = STEP(name="Build",
+            command=["python", build_utils.build_script("build.py")])
+  s2 = STEP(name="Run Tests",
+            command=["python", build_utils.build_script("run_tests.py"), "-f"])
+  s3 = STEP(name="Type Check",
```
rchen152: I'd switch the RunTests and Type Check steps, since we think of type-checking as part of the build/compile step.

build_scripts/travis_script.py
```diff
@@ -0,0 +1,64 @@
...
+def _ismod(f):
+  return f.endswith(".py") and not f.endswith("_test.py")
+
+
+STEP = collections.namedtuple("STEP", ["name", "command"])
+
+PYTYPE_COMMAND = os.path.join("out", "bin", "pytype")
+PYC_DIR = os.path.join(build_utils.PYTYPE_SRC_ROOT, "pytype", "pyc")
+PYC_FILES = [os.path.join(PYC_DIR, f) for f in os.listdir(PYC_DIR) if _ismod(f)]
+TYPECHECK_FILES = [
```
rchen152: The information about which files to type-check is only living here temporarily until we add the functionality that will allow us to put it in setup.cfg, so I think it would be better to keep hard-coding `pytype/compat.py pytype/debug.py pytype/utils.py pytype/pyc --exclude pytype/pyc/*_test.py` in the pytype command rather than trying to turn it into something more maintainable long-term: the former is easier to read, and this won't be around in the long-term anyway.

build_scripts/travis_script.py
martindemello: the original code has typecheck first, and then run the tests if typechecking succeeded.

build_scripts/travis_script.py
```diff
@@ -0,0 +1,64 @@
...
+    "pytype/compat.py", "pytype/debug.py", "pytype/utils.py"] + PYC_FILES
+
+
+def _begin_step(s):
+  print("")
```
martindemello: i think just print() works, no need for the empty string

build_scripts/travis_script.py
sivachandra: I agree with that line of thought if we are using a "different" compiler. If we are using the in-tree compiler, we should first ensure that the compiler works before we use it. That is why I changed the order.

build_scripts/travis_script.py
```diff
@@ -0,0 +1,64 @@
...
+
+from __future__ import print_function
+
+import collections
+import os
+import sys
+
+import build_utils
+
+def _ismod(f):
```
martindemello: what is _ismod? probably needs a more descriptive name.

build_scripts/travis_script.py
sivachandra: I prefer to keep "--exclude pytype/pyc/*_test.py" as well. However, it was working earlier because the shell does the * expansion. Without the help from shell, this is what I could come up with.

build_scripts/travis_script.py
sivachandra: I explained the reason for this above.

build_scripts/travis_script.py
sivachandra: [Just leaving a note for now as it might change based on how the other comments go.]

I wanted a short name so that it does force the line on which it is used to be split into multiple lines.

build_scripts/travis_script.py
martindemello: see if passing shell=True to run_cmd fixes it

rchen152: That seems reasonable to me.

martindemello: i wonder if we should add --exclude_pattern to pytype, or if that would just complicate the interface

build_scripts/travis_script.py
rchen152: Ah, that makes sense. I don't like that the steps feel out of order, but it does seem like the most sensible thing to do.

build_scripts/travis_script.py
rchen152: There's already an --exclude, but it's currently a command-line-only argument. It's on my TODO list to add it to the file config so that we can move exclude patterns to setup.cfg.

martindemello: --exclude takes files and directories as args, and counts on the shell for globbing; exclude_pattern would take a literal glob pattern and filter out files within analyze_project. potentially a bad idea because you'd have to remember to protect against shell expansion when you pass it in, but as a setup.cfg entry it would work fine.

rchen152: Oh, I see. Rather than having both an --exclude and an --exclude_pattern, I think a more user-friendly thing to do would be to treat an exclude= entry in setup.cfg as a glob pattern under the hood, but I suppose what we end up doing will also depend on what's feasible implementation-wise.

martindemello: the issue is for things like `--exclude=*_test.py`, do we want to exclude `**/*_test.py` or just `*_test.py` files in the current directory? the former is probably more intuitive, but then it will do the wrong thing if bash expands the `*`. (it's a common problem i run into when running commands like `find . -name *.py`)

rchen152: Ah, I see what you mean now. Not sure whether it's better to avoid proliferation of config options or to avoid making the user tediously list a bunch of similar excludes, but I've updated the bug to take this into account.

PR #157, opened 2018-09-18
Google sync

rchen152:


rchen152:
Closing because I want to open another one including the parser change.

PR #158, opened 2018-09-18
Google sync

rchen152:


rchen152:
The Travis checks finally completed! They take about half an hour to run now, and both of them failed on the first attempt =S

PR #176, opened 2018-10-03
Google sync

martindemello:


sivachandra:
I think we should add enums.pytd to the CMake targets?

martindemello:
yeah, that looks like the issue.

Issue #173, opened 2018-09-27
have pytype generate ninja files


rchen152:
Generating a ninja file from pytype would solve several problems like parallel execution and rerunning only on changed files, that are handled by the build system internally but not at all in the open source world.

Depending on ninja is not a big problem since it's pip-installable.

rchen152:
Started looking into this a bit. Given a file `foo.py` that imports `bar.py`, the following `build.ninja` file appears to ensure that pytype always analyzes the files in the right order:
```
rule pytype
  command = pytype-single -P pytype_output -o $out $in
build pytype_output/foo.pyi: pytype foo.py | pytype_output/bar.pyi
build pytype_output/bar.pyi: pytype bar.py
```

martindemello:
yay!!

PR #189, opened 2018-11-01
Have pytype (naively) generate ninja files.

### LINKED ISSUE #169: pytype-all should stop analyzing a cycle if all source files have been checked.
As the title says - if the final source file is in a cycle, pytype-all continues doing unnecessary work after the file has been checked. Fix should be as easy as tracking seen filenames and returning early in yield_sorted_modules(): https://github.com/google/pytype/blob/master/pytype/tools/analyze_project/pytype_runner.py

rchen152:
Has pytype farm its pytype-single calls out to ninja via subprocess. At the end
of the change description, I've included a short explanation of how ninja works;
for more detail, there's https://ninja-build.org/manual.html.

I've called this change "naive" because it does not:
(1) avoid clearing pytype_output/ so that output can be reused between runs or
(2) generate fine-grained dependency info so that ninja can parallelize builds.
These improvements will be in future CLs.

Makes progress toward #173. Fixes #169.

* Adds this change to a new CHANGELOG file. We discussed starting a changelog
  in the meeting, so we might as well start now. I've added a single entry for
  the most recent release, to show how the file will be structured; I'll
  backfill the rest of the file later.
* Adds ninja to install_requires in setup.py.
* Moves pyi files into a subdirectory of pytype_output/ so we can have two
  pyi dirs, `pyi` and `pyi_1`.
* Adds a write_ninja_preamble method, which writes out the pytype-single rules
  to build.ninja.
* Adds a write_build_statement method, which appends a single build statement
  to build.ninja.
* Adds a setup_build method, which writes out a full build.ninja file using the
  above two helpers. This method differs in two main ways from the old loop
  over yield_sorted_modules() in PytypeRunner.run(): it skips useless actions,
  and it doesn't count builtin and system files in the progress message. The
  second change was made so that pytype's file count would be more in line
  with ninja's action count.
* Adds a build method, which calls out to ninja via subprocess.

How ninja works:

ninja looks for a build.ninja containing a number of rules and build statements.
For example:

rule infer
  command = pytype-single -o $out $in
build foo.pyi: infer foo.py
build bar.pyi: infer bar.py | foo.pyi

defines an `infer` rule and statements to build `foo.pyi` and `bar.pyi`. In the
second build statement, `| foo.pyi` says that bar.pyi depends on foo.pyi. ninja
will ensure both foo.pyi and bar.pyi are present and up-to-date by executing any
build statements necessary.

The variables $out and $in are provided by default; it's also easy to define
custom ones:

rule infer
  command = pytype-single -P $pythonpath -o $out $in
build foo.pyi: infer foo.py
  pythonpath = pytype_output/pyi/

TESTED=new and existing unit tests, plus pip installing pytype with this change
and running it over some GitHub repos.
PiperOrigin-RevId: 219702660

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,0 +1,5 @@
+Version (upcoming)
+* Generate a build.ninja file and call pytype-single via ninja.
+
+Version 2018.10.30
+* Improve the display of tracebacks in error messages.
--- a/pytype/tools/analyze_project/pytype_runner.py
+++ b/pytype/tools/analyze_project/pytype_runner.py
@@ -4,12 +4,9 @@ from __future__ import print_function
 
 import logging
 import os
-import sys
+import subprocess
 
-from pytype import config as pytype_config
-from pytype import debug
 from pytype import file_utils
-from pytype import io
 from pytype import module_utils
 from pytype.tools.analyze_project import config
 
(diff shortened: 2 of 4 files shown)


rchen152:
FYI, I'm planning to push to GitHub pretty frequently while working on the ninja issue, since that makes development and testing easier.

PR #190, opened 2018-11-06
Google sync

rchen152:


rchen152:
Oh, hmm. `pytype.tools.pyi_checker.definitions_test` failed in all host Python versions except 3.7 - which makes sense, since it uses dataclasses. It looks like xref got around this by simply not providing a `CMakeLists.txt`. I'm going to close this pull request and go sort this out.

Issue #208, opened 2018-12-16
python 3.7 travis tests are using target python_version = (3, 6)


byllyfish:
I'm making a PR to fix issue #203 (import asyncio fails under 3.7). I was puzzled that this failure was not being caught already; there are a few tests that import asyncio.

In pytype/tests/test_base.py, the python_version for Python3 is hard-wired to (3, 6). So, I changed this to use (3, 7), when the pytype host is 3.7. (see https://github.com/google/pytype/compare/master...byllyfish:asyncio_tasks)

This change forces the asyncio failure, but it also elicits a number of other test failures. Here's an example:

```
>>> FAIL - pytype.tests.test_dict - /root/pytype/out/pytype/tests/test_dict.log
```

There are 5 failures in this log. test_pop is one of them.

```
FAIL: pytype.tests.test_dict.DictTest.testPop_py3
Traceback (most recent call last):
  File "/root/pytype/out/pytype/tests/test_base.py", line 536, in Replacement
    return actual_method(self, *args, **kwargs)
  File "/root/pytype/out/pytype/tests/test_dict.py", line 22, in testPop
    """)
  File "/root/pytype/out/pytype/tests/test_base.py", line 459, in assertTypesMatchPytd
    self.assertMultiLineEqual(pytd_tree_src, ty_src)
AssertionError: 'from typing import Dict\n\nd: Dict[str, int]\nv1: int\nv2: None\n' != 'from typing import Any, Dict\n\nd: Dict[str, int]\nv1: Any\nv2: Any\n'
- from typing import Dict
+ from typing import Any, Dict
?                    +++++

  d: Dict[str, int]
- v1: int
?     ^ ^
+ v1: Any
?     ^ ^
- v2: None
+ v2: Any

```

I can confirm this difference using pytype. (It looks like many of the 3.7 test failures are the inferencer punting to Any).

```sh
$ cat test_pop.py 
d = {"a": 42}
v1 = d.pop("a")
v2 = d.pop("b", None)

$ pytype -V 3.7 test_pop.py 
...
$ cat pytype_output/pyi/test_pop.pyi 
# (generated with --quick)

from typing import Any, Dict

d: Dict[str, int]
v1: Any
v2: Any
$ pytype -V 3.6 test_pop.py 
...
$ cat pytype_output/pyi/test_pop.pyi 
# (generated with --quick)

from typing import Dict

d: Dict[str, int]
v1: int
v2: None

```

rchen152:
[...]

PR #217, opened 2018-12-19
Add testing framework for target Python 3.7.

rchen152:
Fixes https://github.com/google/pytype/issues/208.
I tried to keep this change small by skipping tests rather than fixing them, but I did have to implement LOAD_METHOD and CALL_METHOD (or the majority of the 3.7 tests would've failed) and fix a crash involving function.Args.namedargs.

### DIFF
--- a/pytype/tests/test_utils.py
+++ b/pytype/tests/test_utils.py
@@ -443,4 +443,13 @@ def skipUnless37Available(f):
     else:
       _IS_37_AVAILABLE = True
   return unittest.skipUnless(_IS_37_AVAILABLE, "no python3.7")(f)
+
+
+def skipInVersion(python_version, reason):
+  """Skip the test in the (major, minor) version."""
+  def decorate(f):
+    # See test_base.main for how this attribute is used to skip the test.
+    f.__pytype_skip__ = (python_version, reason)
+    return f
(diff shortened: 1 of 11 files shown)


rchen152:
Forgot to add - this PR will have to be imported and re-exported rather than directly merged. I made this change externally because of the related GitHub issue.

pytype/tests/test_utils.py
```diff
@@ -418,3 +421,35 @@ class Py3Opcodes(object):
...
+    try:
+      subprocess.call(["python3.7", "-V"])
+    except OSError:
+      _IS_37_AVAILABLE = False
+    else:
+      _IS_37_AVAILABLE = True
+  return unittest.skipUnless(_IS_37_AVAILABLE, "no python3.7")(f)
+
+
+def skipInVersion(python_version, reason):
```
martindemello: we should never need to skip a test in exactly one version, unless that version is (3, 7). so i'd suggest either hardcoding this to skipIn37 or (if you want something more general) changing it to skipInVersionsFrom so that it automatically skips everything >= the supplied version.

rchen152: Changed to skipIn37.

PR #221, opened 2018-12-23
Travis CI: Use flake8 to find syntax errors and undefined names

cclauss:
Several of these are intentional and those lines can be marked with the linter directive # noqa

__E901,E999,F821,F822,F823__ are the "_showstopper_" [flake8](http://flake8.pycqa.org) issues that can halt the runtime with a SyntaxError, NameError, etc. Most other flake8 issues are merely "style violations" -- useful for readability but they do not effect runtime safety.
* F821: undefined name `name`
* F822: undefined name `name` in `__all__`
* F823: local variable name referenced before assignment
* E901: SyntaxError or IndentationError
* E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 os: linux
 dist: xenial  # Python3.7 virtualenv is available only on Travis Xenial VMs.
-sudo: enabled
 
 git:
   submodules: true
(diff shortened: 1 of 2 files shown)


martindemello:
thanks for the change! you'll need to exclude some directories that contain deliberately broken test data, or python3-only code; the invocation should be
```
flake8 pytype --count --select=E901,E999,F821,F822,F823 --show-source --statistics --exclude=pytype/test_data,pytype/tools
```

.travis.yml
```diff
@@ -1,6 +1,5 @@
 os: linux
 dist: xenial  # Python3.7 virtualenv is available only on Travis Xenial VMs.
-sudo: enabled
```
martindemello: why this change?

cclauss: [Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration)

martindemello: ah, thanks! i missed that entirely.

cclauss:
Modified __.travis.yml__ to add $EXCLUDE and modified __compat.py__ to follow Python porting best practice [___use feature detection instead of version detection___](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection) which placates both PyLint and flake8.

Issue #228, opened 2019-01-17
Pytype does not work with Python distributions like Anaconda


jjedele:
Commonly used Python distributions like Anaconda modify the version string emitted by `python -V` in a way that break the current parsing logic in _utils.py_.

Example:

```
$ python -V
Python 3.6.4 :: Anaconda custom (64-bit)

$ pytype src/add.py
Computing dependencies
Analyzing 1 sources with 0 local dependencies
ninja: Entering directory `/Users/jeff/Downloads/coala-tutorial/pytype_output'
[1/1] pytype-single --imports_info /Users/jeff/Downloads/coala-tutorial/pyt...-annotated --nofail --quick /Users/jeff/Downloads/coala-tutorial/src/add.py
FAILED: /Users/jeff/Downloads/coala-tutorial/pytype_output/pyi/add.pyi
pytype-single --imports_info /Users/jeff/Downloads/coala-tutorial/pytype_output/imports/add.imports --module-name add -V 3.6 -o /Users/jeff/Downloads/coala-tutorial/pytype_output/pyi/add.pyi --analyze-annotated --nofail --quick /Users/jeff/Downloads/coala-tutorial/src/add.py
Traceback (most recent call last):
  File "/Users/jeff/anaconda/bin/pytype-single", line 11, in <module>
    sys.exit(main())
  File "/Users/jeff/anaconda3/lib/python3.6/site-packages/pytype/main.py", line 66, in main
    options = config.Options(sys.argv[1:])
  File "/Users/jeff/anaconda3/lib/python3.6/site-packages/pytype/config.py", line 48, in __init__
    Postprocessor(names, options, self).process()
  File "/Users/jeff/anaconda3/lib/python3.6/site-packages/pytype/config.py", line 357, in process
    node.processor(value)
  File "/Users/jeff/anaconda3/lib/python3.6/site-packages/pytype/config.py", line 481, in _store_python_exe
    python_exe_version = utils.get_python_exe_version(python_exe)
  File "/Users/jeff/anaconda3/lib/python3.6/site-packages/pytype/utils.py", line 158, in get_python_exe_version
    return split_version(python_exe_version.split()[-1].rsplit(".", 1)[0])
  File "/Users/jeff/anaconda3/lib/python3.6/site-packages/pytype/utils.py", line 47, in split_version
    return tuple(map(int, version_string.split(".")))
ValueError: invalid literal for int() with base 10: '(64-bit)'
ninja: build stopped: subcommand failed.
```

[...]

PR #229, opened 2019-01-17
#228 Fix parsing of Python version string.

jjedele:
Python distributions like Anaconda modify the Python version
string in a way that breaks the current parsing logic. This
commit makes the parsing more robust by using regular expressions.

I opened #228 regarding this problem.

### DIFF
--- a/pytype/utils.py
+++ b/pytype/utils.py
@@ -149,13 +149,35 @@ def get_python_exe(python_version):
 
 
 def get_python_exe_version(python_exe):
+  """Determine the major and minor version of given Python executable.
+  
+  Arguments:
+    python_exe: absolute path to the Python executable
+  Returns:
+    Version as (major, minor) tuple"""
   try:
     python_exe_version = subprocess.check_output(
         python_exe + " -V", shell=True, stderr=subprocess.STDOUT).decode()
   except subprocess.CalledProcessError:
     return None
-  # Turn "Python major.minor.micro" into (major, minor)
-  return split_version(python_exe_version.split()[-1].rsplit(".", 1)[0])
+
+  return parse_exe_version_string(python_exe_version)
+
+
+def parse_exe_version_string(version_str):
+  """Parse the version string of a Python executable.
+  
+  Arguments:
+    version_str: Version string as emitted by running `PYTHON_EXE -V`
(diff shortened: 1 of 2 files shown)


pytype/utils.py
```diff
@@ -149,13 +149,35 @@ def get_python_exe(python_version):
...
+
+def parse_exe_version_string(version_str):
+  """Parse the version string of a Python executable.
+  
+  Arguments:
+    version_str: Version string as emitted by running `PYTHON_EXE -V`
+  Returns:
+    Version as (major, minor) tuple"""
+  # match the major.minor part of the version string, ignore the micro part
+  matcher = re.search(r'Python (\d+\.\d+)\.\d+', version_str)
```
jjedele: I see two options here:

1. Parse the major, minor part and continue the already existing `split_version` function.
2. Directly extract the major and minor parts using the regex.

I opted for continuing to use the existing split_version code, but I'm happy to change to option 2 if you prefer that.

jjedele:
I signed it!

Regards,
Jeff

kamahen:
I think that this works for all versions:

```bash
$ python -c 'import sys; print(tuple(sys.version_info))'
(2, 7, 12, 'final', 0)
````

**Comment by jjedele** [comment]
@kamahen : Thanks for your input! You're right that this actually seems to be a more systematic way of doing things.

We could e.g. also use

```
$ python -c "import sys; print(','.join(str(el) for el in sys.version_info))"
3,6,4,final,0
```

This would make the command line call a little messy but the parsing super trivial. Usually I would be all for this as it is probably to be the most robust version. In this case I'm 50:50 since the current version seems a little easier and also robust enough.

@rchen152 : What's your opinion on this?

If we want to change, I can do it later.

rchen152:
@kamahen Thanks for the suggestion!
I'm leaning towards keeping the current version for two reasons:
* The proposed version feels more heavyweight/complex to me, with only a slight benefit.
* The proposed version is considerably slower on my machine. (I ran `python2.7 -V` and `python2.7 -c 'import sys; print(tuple(sys.version_info))'` as subprocess calls 1000x each; the first took 4.8s, the second 17.5s.)

PR #250, opened 2019-03-01
Google sync

rchen152:
First of two GitHub exports. This one is necessary to get the GitHub state to match what the new Copybara config expects; then I'll do a another to export the second MANIFEST.in change.


rchen152:
Ping? (I'm hoping to have enough time to do a second export today.)

PR #261, opened 2019-03-15
Google sync

rchen152:


rchen152:
Ping? I'd like to do a PyPI release today. (On a side note, I wonder if we should stop requiring that copybara PRs go through review, since all the code has already been reviewed once anyway.)

PR #262, opened 2019-03-16
Google sync

rchen152:


rchen152:
Whoops, looks like my variable annotation change broke some target 3.7 tests.

Issue #270, opened 2019-03-26
`python_version` in generated pytype.cfg is different from its python version


tamuhey:
Thanks for this great tool!

I use pytype with virtualenv, and the python version is 3.7:
```
$ python --version
Python 3.7.2
```
and install pytype with pip:
```
$ python -m pip install pytype
```
and generate pytype.cfg:
```
$ pytype --generate-config pytype.cfg
```
but the `python_version` field of `pytype.cfg` is not 3.7, which is the version of venv Python:
```pytest.cfg
python_version = 3.6
```

Is it possible that make the  `python_version` field same as its venv Python version when `pytype.cfg` is generated?

rchen152:
Thanks for using pytype! Looks like we currently have the target python version hard-coded here: https://github.com/google/pytype/blob/fddbad8c00a08b5d90acbbea8b25f7407ae619c7/pytype/tools/analyze_project/config.py#L55. I agree that it would be better to use the venv Python version instead; we should be able to get that with `sys.version_info[:2]`.

tamuhey:
@rchen152 I wrote PR! #272

rchen152:
@tamuhey Thanks for fixing this! The fix has been released in pytype version 2019.03.27.

PR #272, opened 2019-03-27
made `python_verision` in pytype.cfg same as its runtime python version

tamuhey:
from #270

Now the field "python_version" in pytype.cfg is same as its runtime Python version.

Before this PR, always the filed equals "3.6" regardless of what version of Python you use: 
```pytype.cfg
python_version = 3.6
```

After this PR, the fileld equals the version of Python which you use:
```pytype.cfg
python_version = (the version of Python which you use)
```

This is done by using `sys.version_info`.

### DIFF
--- a/pytype/tools/analyze_project/config.py
+++ b/pytype/tools/analyze_project/config.py
@@ -52,7 +52,8 @@ ITEMS = {
         '', '.', None,
         'Paths to source code directories, separated by %r.' % os.pathsep),
     'python_version': Item(
-        '', '3.6', None, 'Python version (major.minor) of the target code.'),
+        '', "{}.{}".format(*sys.version_info[:2]),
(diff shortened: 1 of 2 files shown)


tamuhey:
@googlebot I signed it!

rchen152:
Thank you for your contribution! It looks like you've accidentally added some .pyc files; could you remove those? Otherwise looks good.

tamuhey:
@rchen152 sorry, removed!

rchen152:
Thanks! I'm going to approve this PR, but please don't merge it directly - I have to go through some extra steps to keep the Google-internal and external code bases in sync.

rchen152:
This change has been committed as 43d4a5b8452248d0e308864def54692dc16a717b.

Issue #278, opened 2019-03-30
Use pylint


rchen152:
It would be nice to be able to directly run pylint on PRs rather than importing them into Google and then finding out whether they contain lint errors. We can add a .pylintrc file and use options like
```
[FORMAT]
indent-string='  '
```
to get the right configuration.

hiyashichuka:
I'd like to take this issue.

rchen152:
Absolutely, we'd love your help! A couple thoughts:
* The pylintrc files of other Google open source projects (for example, [Python Fire](https://github.com/google/python-fire/blob/master/pylintrc)) would probably be a useful reference for figuring out a good configuration.
* I suspect some of our code won't lint cleanly no matter what we do, thanks to differences between the Google linter and vanilla pylint, so we may have to disable a lot of error classes.

hiyashichuka:
Thanks for letting me know.
I have two questions I want to ask.

1.
>I suspect some of our code won't lint cleanly no matter what we do, thanks to differences 
>between the Google linter and vanilla pylint, so we may have to disable a lot of error 
>classes.

Examples follow
   vanilla pylint Errors occur.           →C0102: Black listed name "%s"　 judged that as Error
   Google linter  Errors NOT occur. →C0102: Black listed name "%s"　 judged that as NOT Error

In this case,I write this.
[MESSAGES CONTROL]
disable=C0102

Is this understanding of mine correct?

2. Let me know how to execute Google linter.
I would like to confirm differences between the Google linter and vanilla pylint.

Please reply when you have a time.

rchen152:
1. Yep, that's exactly what I was imagining.
2. Unfortunately, the Google linter (gpylint) isn't opensourced. I'd be more than happy to work with you to address any issues that come up, but I don't think there's a way for you to run gpylint directly.

hiyashichuka:
Thanks for letting me know.

I will do as follow. Would this be okay?

I will write pylintrc.
The purpose of the pylintrc is to disable vanilla pylint error. 

If my pylintrc is against the  Google linter, please let me know.

rchen152:
Sounds good to me.

rchen152:
@hiyashichuka Thanks again for fixing this issue!

PR #287, opened 2019-04-04
Add pylintrc #278

hiyashichuka:
I write pylintrc.
The purpose of the pylintrc is to disable vanilla pylint error.

If my pylintrc is against the Google linter, please let me know.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -1,0 +1,158 @@
+[MASTER]
+# Add <file or directory> to the black list. It should be a base name, not a
+# path. You may set this option multiple times.
+ignore=test_data
+
+[MESSAGES CONTROL]
+
+# Disable the message, report, category or checker with the given id(s). You
+# can either give multiple identifier separated by comma (,) or put this option
+# multiple time (only on the command line, not in the configuration file where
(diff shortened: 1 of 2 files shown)


pylintrc
```diff
@@ -0,0 +1,269 @@
...
+# given file (report RP0402 must not be disabled)
+import-graph=
+
+# Create a graph of external dependencies in the given file (report RP0402 must
+# not be disabled)
+ext-import-graph=
+
+# Create a graph of internal dependencies in the given file (report RP0402 must
+# not be disabled)
+int-import-graph=
```
martindemello: remove all the entries that are just set to nothing (and any sections that end up being empty)

pylintrc
```diff
@@ -0,0 +1,269 @@
...
+dummy-variables-rgx=\*{0,2}(_$|unused_|dummy_)
+
+# List of additional names supposed to be defined in builtins. Remember that
+# you should avoid to define new builtins when possible.
+additional-builtins=
+
+[BASIC]
+
+# List of builtins function names that should not be used, separated by a comma
+bad-functions=map,filter,apply,input,reduce
```
martindemello: remove map and filter from this, we use them in pytype.

pylintrc
```diff
@@ -0,0 +1,269 @@
+[MASTER]
```
martindemello: this section can probably be removed

hiyashichuka:
> Thanks for your contribution! Could you also add pylint to .travis.yml so that it gets run regularly?

OK
Shall pylint run all .py file?

rchen152:
All .py files except the ones in test_data, I think.

pylintrc
```diff
@@ -0,0 +1,157 @@
...
+# not require a docstring
+no-docstring-rgx=(__.*__|main|test.*|.*Test|Module)
+
+# Minimum length for a docstring
+docstring-min-length=10
+
+[MISCELLANEOUS]
+
+# List of note tags to take in consideration, separated by a comma.
+notes=FIXME,XXX,TODO
```
rchen152: pytype only uses `TODO` notes, so if you remove `TODO` from this line, then you can also remove `fixme` from the list of disabled messages.

pylintrc
```diff
@@ -0,0 +1,157 @@
...
+
+[FORMAT]
+
+# Maximum number of characters on a single line.
+max-line-length=100
```
rchen152: We use a max line length of 80 but allow exceptions for structured pytype comments and urls. I'd suggest:
```
max-line-length=80
ignore-long-lines=(^.{1,80}\s\s# pytype:\s)|(^\s*# http)
```
Note that this catches a real lint error in merge_pyi.py, which I'm planning to push a fix for today, so you'll have to rebase your branch on top of master.

hiyashichuka:
Thank you for pointing it out.
I corrected the problem points.

PR #295, opened 2019-04-12
Re-enable some disabled pylint messages.

rchen152:


rchen152:
ping?

pylintrc
```diff
@@ -139,7 +125,7 @@ inlinevar-rgx=^[a-z][a-z0-9_]*$
 good-names=i,j,k,ex,main,Run,_,UnpackUnion,MakeClassOrContainerType,Concat,JoinTypes,ast
 
 # Bad variable names which should always be refused, separated by a comma
-bad-names=foo,bar,baz,toto,tutu,tata
+bad-names=toto,tutu,tata
```
martindemello: this seems like a pretty meaningless check. are there any actual bad names we're trying to keep out?

rchen152: Good point, I've removed the remaining bad names. (We can't delete the line altogether, since then pylint would revert to its default list of bad names.)

Issue #90, opened 2018-06-18
PATH issues when running in windows


martindemello:
Working:
- pip install pytype

Not working:
````
> pytype
# (doesn't launch python)

> python c:\...\Python36\Scripts\pytype t.py
cannot run pytype

> python c:\...\Python36\Scripts\pytype-single t.py
cannot find a valid python2.7 interpreter in path
````

In general, the way pytype calls subprocesses does not work in windows. A further issue is that pip install doesn't automatically put the scripts/ directory in PATH, even when running in a virtualenv.

**Comment by rchen152** [comment]
A `build.ninja` issue:

OS: Windows 10
Python version: 3.7
```
D:\New>pytype foo.py
Computing dependencies
Analyzing 1 sources with 0 local dependencies
ninja: error: build.ninja:5: expected build command name
build D:\New\pytype_output\pyi\foo.pyi: check D:\New\foo.py
        ^ near here
ninja: Entering directory `D:\New\pytype_output'
```

hiyashichuka:
Hello 
I'm in the middle of investigating this issue.
I would like to share 2 problems.

1 change python_exe to "python"  

pytype/util.py def get_python_exe_version(python_exe):
` try:
    python_exe_version = subprocess.check_output(
         python_exe  + " -V", shell=True, stderr=subprocess.STDOUT).decode()`   ← change python_exe to "python" 

2.wrong src
Although it is unknown whether this "src" was the direct cause or not, "src" information is wrong.
Src has become 
pytype/pytd/builtins/2/typing.pytd 
or pytype/pytd/builtins/2and3/typing.pytd
or pytype/pytd/builtins/3/typing.pytd

`def parse_string(src, name=None, filename=None, python_version=None,←wrong src
                 platform=None):
  return _Parser(version=python_version, platform=platform).parse(
      src, name, filename)`


However, I cannot solve this issue even though I resolve 2 problem.
I will continue to survey this issue.

rchen152:
I spent some time this weekend trying to install and run pytype on Windows. Didn't get it completely working and gave up after a while, but posting what I did in case it's useful to other people:

**For running pytype**
1. Downloaded Python for Windows from https://www.python.org/downloads/windows/.
1. Downloaded cmake from https://cmake.org/download/.
1. Downloaded git from https://git-scm.com/download/win.
1. Downloaded Microsoft C++ Build Tools from https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2019.

[...]

PR #299, opened 2019-04-22
WIP FIX:PATH issues when running in windows #90

hiyashichuka:
Hi
I have experimentally implemented a Windows support.
OS: Windows 8
Python version: 3.7

1. change python_exe to "python"
(pyc/pyc.py, pytype/utils.py)

2. replace `r` for ``
(pyi/parser.py)

3. change C: to C$:
(pytype_runner.py)

This PR is still work in progress. I have  I think there are some discussion points as a follow:
・`if sys.platform == win32` are frequently.
Should I write proper settings  for each platform in set.py?

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ Platform support:
 * Pytype is currently developed and tested on Linux, which is the main supported
   platform.
 * Installation on MacOSX requires OSX 10.7 or higher and Xcode v8 or higher.
-* Windows is currently not supported.
+* Windows, pytype requires Python 3.3+ to analyze a different Python version from the one it's running under, and also report an error in the code when the launcher isn't found.
 
 ## Installing
 
@@ -199,7 +199,6 @@ of its dependencies.
 
 ## Roadmap
 
-* Windows support
 
 ## License
 Apache 2.0
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ Platform support:
 * Pytype is currently developed and tested on Linux, which is the main supported
   platform.
 * Installation on MacOSX requires OSX 10.7 or higher and Xcode v8 or higher.
-* Windows, pytype requires Python 3.3+ to analyze a different Python version from the one it's running under, and also report an error in the code when the launcher isn't found.
+* Windows requires Python 3.3 or higher
 
 ## Installing
 
--- a/pytype/pyc/pyc.py
+++ b/pytype/pyc/pyc.py
@@ -6,7 +6,6 @@ import re
 import subprocess
 import tempfile
 import sys
-import logging
 from pytype import pytype_source_utils
 from pytype import utils
 from pytype.pyc import loadmarshal
--- a/pytype/pyc/pyc.py
+++ b/pytype/pyc/pyc.py
@@ -6,7 +6,7 @@ import re
 import subprocess
 import tempfile
 import sys
-
+import logging
 from pytype import pytype_source_utils
 from pytype import utils
 from pytype.pyc import loadmarshal
@@ -83,7 +83,10 @@ def compile_src_string_to_pyc_string(src, filename, python_version, python_exe,
     cmd = exe + ["-E", "-", fi.name, filename or fi.name, mode]
     compile_script_src = pytype_source_utils.load_pytype_file(COMPILE_SCRIPT)
 
-    p = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+    try:
+      p = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+    except subprocess.CalledProcessError as e:
+      logging.error("CalledProcessError error: %s\n", str(e))
     bytecode, _ = p.communicate(compile_script_src)
     assert p.poll() == 0, "Child process failed"
   finally:
--- a/pytype/pyc/pyc.py
+++ b/pytype/pyc/pyc.py
@@ -71,8 +71,7 @@ def compile_src_string_to_pyc_string(src, filename, python_version, python_exe,
     # we spawn an external process.
     if python_exe:
       if sys.platform == 'win32':
-        exe = ["py"]
-        exe.append("-" + ".".join(map(str, python_version)))
+        exe = ["py", "-" + ".".join(map(str, python_version))]
       else:
         # Allow python_exe to contain parameters (E.g. "-T")
         exe = python_exe.split()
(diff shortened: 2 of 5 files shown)


rchen152:
Neat! I think checking `if sys.platform == win32` in various places is fine. @martindemello might also have some thoughts on this.

pytype/pyc/pyc.py
```diff
@@ -76,6 +77,8 @@ def compile_src_string_to_pyc_string(src, filename, python_version, python_exe,
     # We pass -E to ignore the environment so that PYTHONPATH and sitecustomize
     # on some people's systems don't mess with the interpreter.
     cmd = exe + ["-E", "-", fi.name, filename or fi.name, mode]
+    if sys.platform == 'win32':
+      cmd = ["python","-E", "-", fi.name, filename or fi.name, mode]
```
martindemello: i think this would not work for analysing python code of a different version than the host python, right? we should add a comment saying so, and maybe something in the readme about this being a limitation under windows unless the python_exe flag is specified.

hiyashichuka: Thank you @martindemello 

I understand your comment like the following.　Please correct me if I’m wrong.

・Points that should be revised 
This code can not know the difference between python2.x and python3.x.
 
・Your proposal
Write to readme about this being a limitation under windows unless the python_exe flag is specified.
→I think that it is an amazing idea,
  but I think about WindowsPytypeUser must provide a value for python_exe.

・My proposal
I would like to know the difference between python2.x and python3.x.

(An example is  python3.7) 
not working under windows.
`subprocess.check_output("python3.7")`

working under windows.
`subprocess.check_output("py -3.7")`

I would like to use py instead of python.
 
What are your thoughts on that?

rchen152: I think using `py -3.7` is a good idea, although it looks like `py` might not always be available: https://docs.python.org/3/using/windows.html#launcher.

martindemello: one further comment - move the `if win32` check above the `cmd = [...]`, and only change the value of `exe`, since the rest is the same. so something like

````
if win32:
  exe = "py -" + version
else:
  exe = "python" + version
````

Comment (hiyashichuka) [reply pytype/pyc/pyc.py]: Thank you @rchen152 @martindemello 

@rchen152 
Thank you for the information. 
I think that I have two choices, what do you think?

1.Add Except 
I'm still thinking about error message.
```
try:
    subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
except FileNotFoundError as e:
    logging.error("FileNotFoundError error 'py' is not recognized: %s\n", str(e))
 ```
2.I will write to readme about "User need check that the launcher is available under windows".

@martindemello 
I  rewrote my code.

rchen152: I think we should do both of those things: write in the README that in Windows, pytype requires Python 3.3+ to analyze a different Python version from the one it's running under, and also report an error in the code when the launcher isn't found.

hiyashichuka: Thanks
I add Except and modify Readme.

pytype/pyc/pyc.py
```diff
@@ -69,17 +70,23 @@ def compile_src_string_to_pyc_string(src, filename, python_version, python_exe,
     # In order to be able to compile pyc files for both Python 2 and Python 3,
     # we spawn an external process.
     if python_exe:
-      # Allow python_exe to contain parameters (E.g. "-T")
-      exe = python_exe.split()
+      if sys.platform == 'win32':
+        exe = ["py"]
```
rchen152: nit: you can combine this line and the next one as:
`exe = ["py", "-" + ".".join(map(str, python_version))]`

pytype/pyc/pyc.py
```diff
@@ -69,17 +70,23 @@ def compile_src_string_to_pyc_string(src, filename, python_version, python_exe,
...
     # on some people's systems don't mess with the interpreter.
     cmd = exe + ["-E", "-", fi.name, filename or fi.name, mode]
-
     compile_script_src = pytype_source_utils.load_pytype_file(COMPILE_SCRIPT)
 
-    p = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+    try:
+      p = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+    except subprocess.CalledProcessError as e:
+      logging.error("CalledProcessError error: %s\n", str(e))
```
rchen152: I'd raise an error here so that if the subprocess call fails and `p` isn't defined, we don't try to use `p` later on. Maybe an `IOError` would be appropriate.

pytype/pyi/parser.py
```diff
@@ -431,6 +432,10 @@ def parse(self, src, name, filename):
     self._parent_name = module_utils.get_package_name(self._package_name, False)
 
     try:
+      if sys.platform == 'win32':
+        if type(src) is type(bytes):
```
rchen152: nit: `isinstance(src, bytes)` would be more in line with pytype's conventions.

README.md
```diff
@@ -62,7 +62,7 @@ Platform support:
 * Pytype is currently developed and tested on Linux, which is the main supported
   platform.
 * Installation on MacOSX requires OSX 10.7 or higher and Xcode v8 or higher.
-* Windows is currently not supported.
+* Windows, pytype requires Python 3.3+ to analyze a different Python version from the one it's running under, and also report an error in the code when the launcher isn't found.
```
rchen152: With the current implementation, it looks like we always try to use the `py` launcher, so I'd suggest changing this to simply: "Windows requires Python 3.3 or higher."

hiyashichuka:
Thanks @rchen152 
I am always helped by your polite treatment. 

I can't apologize enough.
I noticed My Pull Request not perfect.

My Pull Request has number of problems.
1. There is not .travis.yml for windows
2. Not all behavior are same.
Same piece of code, different behavior on Windows and Linux.

Exsample
```
import re
```
Linux work 
Windows not work

Exsample
```
import os
print(os.getcwd())
```
Linux work 
Windows work

I'd like to resolve this issue, if you  give me time.

rchen152:
@hiyashichuka There's absolutely no need for you to apologize, and please take as much time as you want! We really appreciate your previous contribution to pytype and are excited about Windows support.

hiyashichuka:
Thanks @rchen152 
Could you please let me know what you think about my suggestion?

I  make a list of all the urgent tasks we need to do.
Please close issue  when  I completed fix tasks.

Tasks
1. Create .travis.yml for windows
2. Fix: Same piece of code, different behavior on Windows and Linux.
```
import re
```

There might be a problem but I would like to set the deadline.

rchen152:
That sounds good to me! (By the way, sorry for the slow responses; I'm currently at the PyCon conference.)

hiyashichuka:
> That sounds good to me! (By the way, sorry for the slow responses; I'm currently at the PyCon conference.)

hiyashichuka:
Sorry I made a mistake operating.
I am going to deal with this problem positively.
This may take some time, so have some patience.

hiyashichuka:
Sorry, I left the PR.
I still haven't gotten started on the problem.
May i close the PR?

rchen152:
Of course, no worries.

PR #340, opened 2019-06-08
Google sync

rchen152:


rchen152:
Closing, there's a py3-ism in tests.test_type_comments that needs to be removed.

PR #343, opened 2019-06-17
Update README.md

adamcataldo:
Giving some nice examples of where pytype shines, and making them front and center in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,36 @@ well-typed and error-free.
 
 For more information, check out the [user guide][user-guide] or [FAQ][faq].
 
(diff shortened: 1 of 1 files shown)


rchen152:
Looks good, I'll merge the PR once the tests pass.

kamahen:
You might also want to mention that pytype can infer types, e.g.:
```python
def build_list(x):
    lst = []
    if x > 0:
        lst.append(Foo())
    return lst
```
it infers:
```python
def build_list(x) -> List[Foo]: ...
```

PR #345, opened 2019-06-19
Google sync

rchen152:


rchen152:
Whoops, annotate_ast_test is py2-only >< Will close and try again.

PR #358, opened 2019-07-16
Updating Setup.cfg

priyansh19:
Adding comments in Setup.cfg file

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,9 @@
 [pytype]
+# basic Python version needed to install pytype. you can use 2.7 or latest versions also.
 python_version = 2.7
+
+# Modules To be imported in Setup.py 
+
 inputs =
     pytype/compat.py
     pytype/debug.py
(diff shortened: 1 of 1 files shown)


rchen152:
Hi Priyansh! Thank you for the PR. Could you explain what the purpose of adding these comments is? This setup.cfg file isn't the template that pytype uses to generate configurations; it's pytype's own configuration to type-check itself.

priyansh19:
I thought, it was needed so as to help beginners to understand the basic requirements and libraries that will be used in Pytpe.

KevinHock:
IMO This is a slippery slope, where one could add comments to anything for beginners, e.g.

# iterate through thing
For t in thing:

So this feels somewhat unnecessary.

rchen152:
Sorry for letting this sit for so long. Now that I'm revisiting it, I agree with @KevinHock - these comments don't appear to add much value. I'm going to close this PR, but @priyansh19, thank you for your time on it.

PR #381, opened 2019-08-09
Use README.md for the PyPI description.

rchen152:
As it turns out, reStructuredText is not happy with unicode characters.
But since March 2018, PyPI has allowed using markdown for the long
description, which works: https://test.pypi.org/project/pytype/.


martindemello:
is it safe to delete DESCRIPTION.rst?

rchen152:
I think so! That's tracked internally (urgh), so I made a separate change to delete it.

Issue #383, opened 2019-08-13
When pytype finds an error in a method signature, it prints a confusing error message about calling the method even if it is only declared.


dumblob:
This is just a minor detail, but I'm reporting it as I got very surprised by **was called** as the type checking happened only in code where there were no calls to this method.
```python
class C:
  def hash( x: str ):
    return True
```
```
Computing dependencies
Analyzing 4 sources with 0 local dependencies
ninja: Entering directory `/home/user/app/.pytype'
[1/1] check main
FAILED: /home/user/app/.pytype/pyi/main.pyi 
pytype-single --imports_info /home/user/app/.pytype/imports/main.imports --module-name main -V 3.7 -o /home/user/app/.pytype/pyi/main.pyi --analyze-annotated --nofail --quick /home/user/app/main.py
Function C.hash was called with the wrong arguments [wrong-arg-types]
  Expected: (x: str)
  Actually passed: (x: C)

For more details, see https://google.github.io/pytype/errors.html#wrong-arg-types.
ninja: build stopped: subcommand failed.
```

martindemello:
this is a bit of an odd edge case, and i'm not sure what the right way to fix it is. the underlying issue is that pytype analyses a function by "calling" it, but with the input variables not bound to anything; however, in the case of a method it is implicitly called with the class as the first argument, and that is what triggers the type mismatch, not the definition of `hash` itself. so pytype is technically correct (the best kind of correct!) but that is indeed a confusing error message.

domandinho:
I created PR with suggestion, which will resolve this error, but i am not sure is this correct approach. Please take a look at that:
https://github.com/google/pytype/pull/406

Kind Regards
Piotr Domański

rchen152:
I'm going to close this, since there's no clear way to fix it. pytype analyzes functions by calling them with dummy arguments, so at the point that the error is logged, pytype doesn't know whether the function was called by the code or for analysis purposes.

dumblob:
Well, there is a workaround IMHO. This specific exception could be caught by pytype and prolonged with an explanation, that **pytype actually called it**, but there seem to be some more fundamental issue which pytype couldn't even analyze yet.

PR #406, opened 2019-09-08
Change default value of Error.filename to empty string instead of None

domandinho:


domandinho:
I saw TODO in existing code to disallow None as a filename and I found out that when filename is empty string parser behave differently and doesn't print the "File" section. Making empty string as default value will resolve the problem of sorting errors. It won't resolve a problem with a confusing error message, but there is a separate issue for that (https://github.com/google/pytype/issues/383).

I am not sure is it the right approach and this is only the suggestion of change. I saw a similar approach in the existing code for lineno field, where the default value is set to 0, when unknown.

Please take a look at that.
Kind Regards
Piotr Domański

domandinho:
@googlebot I signed it!

martindemello:
merged #412 instead

PR #412, opened 2019-09-15
Temporary fix for sorting errors

domandinho:
An alternative way to avoid exception during sorting errors.
This can be a temporary fix if you want to remove nullability of fileno field or disallow creating errors without lineno in the future.


martindemello:
thanks! we will merge this into our internal branch and then reexport to github

martindemello:
merged internally, it should go out in the next github sync

PR #431, opened 2019-10-12
Documentation Edits wording/grammer/cleanup

SemicolonExpected:
ln 17 - don't need the meaning, if we just replace it with a semicolon it can be inferred that static analyzers dont execute the code it runs on
ln 28 I think its a bit cleaner to swap the syntax of the sentence so it ends with "like this:" as ending on "which others would miss"  and then presenting the snippet is a bit awkward compared to presenting the snippet after saying "Like this"
ln83 similarly "which replaces" sounds better than replacing

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ optional for pytype, it will check and apply them where present.
 which can be merged back into the Python source with a provided
 [merge-pyi][merge-pyi] tool.
 
-Pytype is a static analyzer, meaning it does not execute the code it runs on.
+Pytype is a static analyzer; it does not execute the code it runs on.
 
 Thousands of projects at Google rely on pytype to keep their Python code
 well-typed and error-free.
@@ -25,7 +25,7 @@ For more information, check out the [user guide][user-guide] or [FAQ][faq].
 
 1. Pytype uses **inference** instead of gradual typing. This means it will
 infer types on code even when the code has no type hints on it. So it can
-detect issues with code like this, which other type checkers would miss:
+detect issues with code which other type checkers would miss like this:
 
     ```python
     def f():
@@ -80,7 +80,7 @@ also easy to add pytype to your automated testing; see this
 
 Finally, pytype generates files of inferred type information, located by default
 in `.pytype/pyi`. You can use this information to type-annotate the
-corresponding source file, replacing `module.py` with the file's import path:
+corresponding source file, which replaces `module.py` with the file's import path:
 
 ```
 merge-pyi -i module.py .pytype/pyi/module.pyi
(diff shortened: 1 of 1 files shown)


SemicolonExpected:
@googlebot I signed it!

README.md
```diff
@@ -25,7 +25,7 @@ For more information, check out the [user guide][user-guide] or [FAQ][faq].
 
 1. Pytype uses **inference** instead of gradual typing. This means it will
 infer types on code even when the code has no type hints on it. So it can
-detect issues with code like this, which other type checkers would miss:
+detect issues with code which other type checkers would miss like this:
```
rchen152: I prefer the original structure here, for the reason that it keeps "like this" close to the word ("code") that the phrase describes.

README.md
```diff
@@ -80,7 +80,7 @@ also easy to add pytype to your automated testing; see this
 
 Finally, pytype generates files of inferred type information, located by default
 in `.pytype/pyi`. You can use this information to type-annotate the
-corresponding source file, replacing `module.py` with the file's import path:
+corresponding source file, which replaces `module.py` with the file's import path:
```
rchen152: This changes the meaning; "replacing ..." is meant to be an instruction to the reader, not a description of what the type-annotating does.

I agree that the sentence is clunky - suggestions welcome on how to improve it.

README.md
SemicolonExpected: I see your point and will revert this one back

README.md
SemicolonExpected: Ok I think I see what you're trying to do now. 
Correct me if I'm wrong but you're saying in the instruction below 

`merge-pyi -i module.py .pytype/pyi/module.pyi`

the parameter `module.py` should be replaced with the file's import path.

rchen152: That's correct.

SemicolonExpected: So I have two suggestions, and one is by saying 'by replacing' which adds an extra word but makes things a lot more clear what is going on, or just make it known in the command that its suppose to be the users file path similar to how you would document a program's parameters

rchen152: Thanks! I'm not entirely sure about calling the parameters filepath and filename, since the second parameter is supposed to be the path corresponding to the module name (e.g., "foo/bar/baz.pyi" for module "foo.bar.baz"), but this is already clearer than what we had before, so I'm going to go ahead and accept this PR and maybe tweak the language later if I think of a way to improve it :)

FYI the process we use for keeping our Google-internal and external code bases in sync will close this PR rather than merging it, but you'll still be credited as the author in the corresponding commit on master.

PR #433, opened 2019-10-13
Update README.md

alanyee:
Add link to LICENSE file


rchen152:
Thanks! This change looks good to me. Due to the process we use to keep the Google-internal and external code bases in sync, this PR will be closed rather than merged, but you'll be credited as the author of the corresponding commit in master.

PR #443, opened 2019-11-01
typing.Collection shouldn't inherit from Container.

Flameeyes:
Despite the documentation
(https://docs.python.org/3/library/collections.abc.html#collections.abc.Collection)
suggesting the class should include the __contains__ method, this is
not actually implemented by (at least) ValuesView (dict_values), which
most definitely _is_ a Collection.

### DIFF
--- a/pytype/pytd/builtins/2/typing.pytd
+++ b/pytype/pytd/builtins/2/typing.pytd
@@ -153,7 +153,7 @@ class Sequence(Sized, Iterable[_T], Container[_T], Reversible[_T], Protocol):
   def __reversed__(self) -> Iterator[_T]: ...
 
 
-class Collection(Sized, Iterable[_T], Container[_T], Protocol):
+class Collection(Sized, Iterable[_T], Protocol):
   __slots__ = []
 
 
--- a/pytype/pytd/builtins/3/typing.pytd
+++ b/pytype/pytd/builtins/3/typing.pytd
@@ -153,7 +153,7 @@ class Sequence(Sized, Iterable[_T], Container[_T], Reversible[_T], Protocol):
   def __reversed__(self) -> Iterator[_T]: ...
 
 
-class Collection(Sized, Iterable[_T], Container[_T], Protocol):
+class Collection(Sized, Iterable[_T], Protocol):
   __slots__ = []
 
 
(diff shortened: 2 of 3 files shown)


rchen152:
ValuesView does have a `__contains__` method, so I think a better fix would be to add `__contains__` here: https://github.com/google/pytype/blob/55dcaa09d1973f17cef60b29392857be46f9fa59/pytype/pytd/builtins/2and3/typing.pytd#L412. (Weirdly enough, dict_values does not have `__contains__`, but based on the typing definition, I think the intent is for dictionary views to be containers.)

Flameeyes:
So this was briefly discussed with @Yhg1s before I sent the pull request. AIUI `ValuesView` is intended to model `dict_values` — and so any differences between the two should be fixed by changing the former.

The problem is that we can't easily fix the two statements:

- `dict_values` is a `ValuesView`;
- `Collection` is a `Container`.

In particular, faking that `dict_values` is a `Container` (by saying that `__contains__` is implemented) might work, but then blurries the line of how correct the typing should be… I'm open to either case.

Note that the other problem is that `Container` expects an explicit `__contains__`, but the latter is not required to implement `in` (which I didn't realize until I found this problem). In particular, `Iterator` classes would allow `in` (but not with a stable behaviour). So right now it feels like the semantics of `Container` would be "It's safe to use the `in` operator." — and in that case, `dict_values` would logically be a `Container` even without `__contains__`.

rchen152:
I don't think we should be guessing at what the definitions should be based on the implementations of these various classes, given that the implementations are - as you pointed out - sometimes confusing and contradictory.

The documentation at https://docs.python.org/3/library/collections.abc.html says that ValuesView is a Collection and has `__contains__` as a mixin method, and that Container is defined as an ABC that provides `__contains__`. The documentation at https://docs.python.org/3/library/typing.html#typing.Collection says that a Collection is a Sized Iterable Container.

The intent seems pretty clearly to be that a class needs to implement `__contains__` to satisfy Container (and therefore Collection) and that ValuesView should have `__contains__`.

rchen152:
@Flameeyes are you still interested in pursuing this PR?

rchen152:
I'm going to close this for now. Feel free to try again.

PR #454, opened 2019-11-18
Add a -j/--jobs option to pytype

dlax:
The option is simply passed to ninja, thus enabling parallel builds.


martindemello:
thanks! we will import and merge this via our internal workflow and sync back to github

PR #457, opened 2019-11-23
Fix UnboundLocalError about 'argument_parser' in Options

dlax:
The except clause in Options.__init__() refers to 'argument_parser'
variable which was only defined in an "if" clause. This results in the
following error:

  UnboundLocalError: local variable 'argument_parser' referenced before assignment

Move 'argument_parser' definition at function's global scope.

### DIFF
--- a/pytype/config.py
+++ b/pytype/config.py
@@ -44,8 +44,8 @@ class Options(object):
     Raises:
       sys.exit(2): bad option or input filenames.
     """
+    argument_parser = make_parser()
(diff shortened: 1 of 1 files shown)


kamahen:
@rchen152  Isn't this a bug that pytype should have caught by flow analysis? (Or, if not pytpye, then pylint?)

rchen152:
Thanks! I restarted a Travis test that looks like it failed due to a download failure (?); I'll import this PR once that passes.

PR #458, opened 2019-11-24
python -m pytype.single

dlax:
First commit allows running "pytype-single" as "python -m pytype.single" following similar change about "pytype" executable 79f8d35d93f8798d3bf7be0120c8a058a8af5280.

Second commit uses this executable to generated ninja files.


rchen152:
This will fail on systems where sys.executable is None (https://docs.python.org/3/library/sys.html#sys.executable). This isn't just a theoretical concern - I imported this PR and discovered that sys.executable appears to be `None` in our internal testing environment :(

I think the quickest (albeit ugly) way to fix this would be to do:
```
if sys.executable:
  [new code]
else:
  [old code]
```
everywhere where this PR changes the code to use sys.executable.

dlax:
@rchen152, I added commit a384401 to address the `sys.executable` issue.

rchen152:
Thanks! I tested this out and it seems to be working as expected. (By the way, since this PR touches setup.py, I'm going to merge it directly and manually copy the changes internally - in case you're wondering why this PR is being merged when the others were closed.)

PR #496, opened 2020-01-21
Supporting multiple disabled rules in .cfg files

lidizheng:
Currently, in .cfg files, there isn't a documentation mentioning how to set multiple disable rules. After experiments, I found the only valid way is:

```
[pytype]
...
disable = "import-error,module-attr"
```

However, users are expecting the normal config file syntax as:

```
[pytype]
...
disable =
  import-error
  module-attr
```

So, this PR added this support. It is backward compatible. Test case included.

If this project is export-only, see the internal version of this change cl/290806072.

### DIFF
--- a/pytype/tools/analyze_project/config.py
+++ b/pytype/tools/analyze_project/config.py
@@ -193,7 +193,7 @@ def generate_sample_config_or_die(filename, pytype_single_args):
   items = dict(ITEMS)
   assert set(_PYTYPE_SINGLE_ITEMS) == set(pytype_single_args)
   for key, item in _PYTYPE_SINGLE_ITEMS.items():
-    if items[key].comment is None:
+    if item.comment is None:
       items[key] = item._replace(default=pytype_single_args[key].default,
                                  comment=pytype_single_args[key].help)
     else:
--- a/pytype/tools/analyze_project/config.py
+++ b/pytype/tools/analyze_project/config.py
@@ -62,7 +62,9 @@ ITEMS = {
 
 # The missing fields will be filled in by generate_sample_config_or_die.
 _PYTYPE_SINGLE_ITEMS = {
-    'disable': Item(None, 'pyi-error', ArgInfo('--disable', ','.join), None),
+    'disable': Item(
+        None, 'pyi-error', ArgInfo('--disable', ','.join),
+        'Comma or space separated list of error names to ignore.'),
     'report_errors': Item(
         None, 'True', ArgInfo('--no-report-errors', lambda v: not v), None),
     'precise_return': Item(
--- a/pytype/tools/analyze_project/config.py
+++ b/pytype/tools/analyze_project/config.py
@@ -86,8 +86,8 @@ def string_to_bool(s):
   return s == 'True' if s in ('True', 'False') else s
 
 
-def concate_disabled_rules(s):
-  return ','.join(t for t in s.split('\n') if t)
+def concat_disabled_rules(s):
+  return ','.join(t for t in s.split() if t)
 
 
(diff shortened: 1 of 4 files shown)


rchen152:
We have documentation in multiple places that disabled error classes need to be comma-separated: it's in our [README](https://github.com/google/pytype#usage), which is also mirrored to our [home page](https://google.github.io/pytype/), and appears as a comment in the sample configuration file produced by `pytype --generate-config`.

I'm not necessarily opposed to supporting whitespace-separated error classes as well, but the PR would need to be amended to update both the docs and the help text for `--disable`. For consistency, it would also make sense to support the same in `pytype-single` as well.

An easier solution would be to make the existing documentation more easily discoverable - where did you look for information on how multiple disables work?

lidizheng:
@rchen152 Documentation updated. PTALA

pytype/tools/analyze_project/config.py
```diff
@@ -62,7 +62,9 @@
 
 # The missing fields will be filled in by generate_sample_config_or_die.
 _PYTYPE_SINGLE_ITEMS = {
-    'disable': Item(None, 'pyi-error', ArgInfo('--disable', ','.join), None),
+    'disable': Item(
+        None, 'pyi-error', ArgInfo('--disable', ','.join),
+        'Comma or space separated list of error names to ignore.'),
```
rchen152: For this to work, you'll also need to change the implementation of `generate_sample_config_or_die` below to not overwrite the comment if it is non-`None`.

pytype/tools/analyze_project/config.py
```diff
@@ -84,6 +86,10 @@ def string_to_bool(s):
   return s == 'True' if s in ('True', 'False') else s
 
 
+def concate_disabled_rules(s):
```
rchen152: nit: `concate` -> `concat`

pytype/tools/analyze_project/config.py
```diff
@@ -84,6 +86,10 @@ def string_to_bool(s):
...
+  return ','.join(t for t in s.split('\n') if t)
```
rchen152: If you do `s.split()` without any arguments, it will split on both `\s` and `\n`, which is slightly nicer.

pytype/tools/analyze_project/config.py
lidizheng: Good catch! Updated `generate_sample_config_or_die`.

pytype/tools/analyze_project/config.py
lidizheng: Done.

pytype/tools/analyze_project/config.py
lidizheng: Done.

pytype/tools/analyze_project/config.py
```diff
@@ -186,8 +193,11 @@ def generate_sample_config_or_die(filename, pytype_single_args):
   items = dict(ITEMS)
   assert set(_PYTYPE_SINGLE_ITEMS) == set(pytype_single_args)
   for key, item in _PYTYPE_SINGLE_ITEMS.items():
-    items[key] = item._replace(default=pytype_single_args[key].default,
-                               comment=pytype_single_args[key].help)
+    if items[key].comment is None:
```
rchen152: This should be `item.comment`, not `items[key].comment`.

lidizheng: Done. Code simplified.

rchen152:
Thanks again for the PR! The content has been merged as 26829d6c0e2f5bff6d091819f1ea3c59d6619c76.

PR #507, opened 2020-02-06
Improvements to setup.py

pganssle:
This is a first pass improvement to `setup.py` to take advantage of newer packaging features and bring this package closer into line with current packaging best practices. The two major changes are:

1. Replace conditional logic with environment markers
    
    PEP 496 introduced environment markers, which allow you to declare a dependency conditional on the environment the package is to be installed into. Among other things, this is useful for cross-compilation.

2. Add python_requires matching test matrix
    
    This will prevent `pip` from finding versions of pytype that do not support a given version of Python. Unfortunately, when no python_requires declaration is made, the default is "acceptable for all versions", and so without yanking old versions, `pip install pytype` on unsupported versions will simply fall back to the latest release without a `python_requires` declaration. Alas, this still seems like a win.

I have also updated the trove classifiers to include Python 3.7 support. I was uncertain why this package is tested on 3.5 but not included in the trove classifiers, so I have left it out of the classifiers but included 3.5 in the `python_requires` declaration.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -167,6 +167,7 @@ setup(
         'License :: OSI Approved :: Apache Software License',
         'Programming Language :: Python',
         'Programming Language :: Python :: 2.7',
+        'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: Implementation :: CPython',
--- a/setup.py
+++ b/setup.py
@@ -167,6 +167,7 @@ setup(
         'Programming Language :: Python',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: Implementation :: CPython',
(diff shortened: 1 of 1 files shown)


rchen152:
Wow, thanks for the PRs! I'll try to do an initial review of them before the end of today.

setup.py
```diff
@@ -165,14 +152,23 @@ def get_install_requires():
...
+        'typed_ast; python_version >= "3.3"',
+    ],
     classifiers=[
         'Development Status :: 5 - Production/Stable',
         'Intended Audience :: Developers',
         'License :: OSI Approved :: Apache Software License',
         'Programming Language :: Python',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
```
rchen152: Could you add 3.5 to this list? I'm pretty sure we just forgot to update it when we decided to support 3.5.

pganssle: Done (and ported this change to #509).

pganssle:
Closed in favor of #509

Issue #376, opened 2019-08-08
Run tests against installed version of Python


pganssle:
This issue was triggered by #372 - it is a fairly common problem that the installed version of a package does not match the version in the repo if the tests are not run against the version *as installed*, which is how the testing is currently working here. This is in fact one of the [major downsides of not using a src/ layout](https://hynek.me/articles/testing-packaging/). It would be a good idea to always test against the library *as installed*, so you can catch this as part of the PR.

To ameliorate this and other issues, I highly recommend running the tests with [`tox`](https://tox.readthedocs.io/en/latest/) `tox`, and possibly switching over to using an src/ layout for the code. Even if you do not switch over to using a `src/` layout, you can use either `changedir={toxworkdir}` or `changedir={envtmpdir}`. There is a somewhat old (and now only available via the Internet Archive) blog post about that [here](https://web.archive.org/web/20170615032800/https://enotuniq.org/).

**Edit**: I have written [a blog post](https://blog.ganssle.io/articles/2019/08/test-as-installed.html) on how to configure `tox` to test against the installed version. It does not include the stuff about how to correctly configure `coverage.py`, but I don't think `pytype` is using `coverage.py` at the moment.

martindemello:
thanks, i wasn't aware of the whole src/ layout discussion, it looks like it has some major upsides. we will take a look at tox too; as you say it would definitely have caught #372 and possibly some earlier setup-related bugs as well.

h-joo:
Hi, thank you for your report. Google is shifting its effort to a different approach for type checking apart from pytype, and we're not planning to put in any effort in the near future in pytype, so we'll close this issue. Please see the announcement [here](https://github.com/google/pytype/blob/main/README.md).

PR #508, opened 2020-02-06
Add __init__ files to two tools modules

pganssle:
These two modules were not declared as modules because their `__init__.py` is missing. Python 3+ compensates for this by considering them namespaces, but the `annotate-ast` command (and probably a few others) are broken on Python 2.7 due to lack of an `__init__.py`

Here is an MWE (it assumes you have `virtualenv` installed):

```
cd $(mktemp -d)
virtualenv venv --python=`which python2.7`
source venv/bin/activate
pip install pytype
hash -r  # Ensure that annotate-ast points to the python2.7 version
annotate-ast a.py
```

Prior to this PR, this should fail with some message like this:

```
$ annotate-ast
Traceback (most recent call last):
  File "/tmp/tmp.VKZgcka8q6/venv/bin/annotate-ast", line 5, in <module>
    from pytype.tools.annotate_ast.main import main
ImportError: No module named annotate_ast.main
```

I have not added a test for this, but I suspect that #376, combined with moving the tests into a dedicated `tests/` directory (rather than in-line, as they are now), will catch this and many other issues like it.

### DIFF
--- a/pytype/tools/annotate_ast/main.py
+++ b/pytype/tools/annotate_ast/main.py
@@ -1,6 +1,8 @@
 # python3
 """Call annotate_ast on a source file."""
 
+from __future__ import print_function
+
 import argparse
 import sys
 


pganssle:
This has apparently started failing on Python 2.7 because the `__init__.py` is triggering `pylint` to actually lint the files, and `pytype.tools.annotate_ast.main` contains a print statement with parens around it. It appears that [the file is intended to be linted as Python 3](https://github.com/google/pytype/blob/a735b1109229cacfcbbf183d1ea85d2c31065d70/pytype/tools/annotate_ast/main.py#L1), but `pylint` doesn't know that, so I think the options are:

1. Add a `from __future__ import print_function` statement at the top of the file.
2. Explicitly ignore the file in [`travis_script.py`](https://github.com/google/pytype/blob/a735b1109229cacfcbbf183d1ea85d2c31065d70/build_scripts/travis_script.py#L45) as is being done for `xref`.

I would lean towards the first one as long as the package supports Python 2 anyway.

rchen152:
Either option sounds fine. 1) is probably simpler, since I vaguely recall having to change multiple files to exclude xref from Python 2 linting.

pganssle:
@rchen152 Great, I went with #1 and it seems to have fixed the linting issue. Thanks!

Issue #372, opened 2019-08-05
pytype setup omits pytype.tools.traces


kamahen:
This is with the latest pull from github (2019.7.30). I fixed it by:
```
cp -pir ./pytype/tools/traces/ ~/.local/lib/python3.6/site-packages/pytype/tools/
```

Here's the error I got:
```
Traceback (most recent call last):
  File "/home/peter/.local/bin/pyxref", line 11, in <module>
    load_entry_point('pytype==2019.7.30', 'console_scripts', 'pyxref')()
  File "/home/peter/.local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/home/peter/.local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2843, in load_entry_point
    return ep.load()
  File "/home/peter/.local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2434, in load
    return self.resolve()
  File "/home/peter/.local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2440, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/home/peter/.local/lib/python3.6/site-packages/pytype/tools/xref/main.py", line 15, in <module>
    from pytype.tools.xref import indexer
  File "/home/peter/.local/lib/python3.6/site-packages/pytype/tools/xref/indexer.py", line 18, in <module>
    from pytype.tools.traces import source
ModuleNotFoundError: No module named 'pytype.tools.traces'
```

kamahen:
I used these commands to install (I had previously deleted all pytype files from ~/.local):
```
cd ~/src/pytype
git pull --recurse-submodules
git submodule init
git submodule update
python3 -m pip install --upgrade --user .
```

rchen152:
Thanks for catching this; I opened a PR to fix it.

kamahen:
Verified fixed.

PR #509, opened 2020-02-06
Further modernize python build system

pganssle:
This is the somewhat more radical set of logical next steps after #507 - things that may be slightly more controversial since they opt in to a more modern build system and don't necessarily fix bugs.

The biggest major changes here are:

- Addition of a `pyproject.toml`, which opts in to builds using [PEP 517](https://www.python.org/dev/peps/pep-0517/) and [PEP 518](https://www.python.org/dev/peps/pep-0518/). I have split this into two commits, the first adds `pyproject.toml`, which declares the backend correctly when installing from the repo. At the moment, `pyproject.toml` is not automatically included in source distributions, so with only this commit, end users installing from `sdist` won't be using the PEP 517 machinery. The second commit adds `pyproject.toml` to the `MANIFEST.in` to fix this. Note that this also requires a minor `sys.path` hack, because the `setuptools` backend no longer includes the repo root on the Python path by default.

- Move everything that's easy to move from `setup.py` into `setup.cfg`.

  Using the declarative build file for all your declarative options is current best practice, since it makes it much easier to parse metadata from projects without executing arbitrary code.

  This will change the minimum required version of setuptools to 30.4, which was released in December 2016.

- Switched over the package listing to use `setuptools.find_packages`, which will make it less likely that you will encounter bugs like #372, but it does mean that you are using a somewhat more complicated whitelist/blacklist to manage the list. I think the trade-off is worth it, and it can be dramatically simplified by switching to a `src/` layout, as I recommend in #376.

This branch has #507 as its head and cherry-picks #508, since this relies on the fix in #508 to avoid those modules being excluded from installation. I think you can merge the other two first and then I'll rebase this against master, or you can merge this directly and close the other two, whatever works for your workflow.

I have tried to separate this into clean and mostly independent commits so that if you don't want to make any given change you can revert it individually.


setup.cfg
```diff
@@ -1,3 +1,60 @@
...
+console_scripts =
+    annotate-ast = pytype.tools.annotate_ast.main:main
+    merge-pyi = pytype.tools.merge_pyi.main:main
+    pytd = pytype.pytd.main:main
+    pytype = pytype.tools.analyze_project.main:main
+    pytype-single = pytype.single:main
+    pyxref = pytype.tools.xref.main:main
+
 [pytype]
 python_version = 2.7
```
martindemello: should we change this while we're at it?

rchen152:
This is much nicer than what we had before =D Didn't notice any issues aside from Martin's comments.

setup.cfg
```diff
@@ -1,3 +1,60 @@
+[metadata]
+name = pytype
+description = Python type inferencer
+version: attr:pytype.__version__.__version__
```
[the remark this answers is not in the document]
pganssle: Whoa. Weirdly, `version:` actually works. INI is a terrible format and support for `pyproject.toml` cannot come to `setuptools` too soon.

setup.cfg
pganssle: I don't really understand this part of it, so if you don't mind I'd rather leave that to someone else.

pganssle:
I've rebased this against `fix_setup_py`, dropping the commits from #508 that were merged, and I've rebased that one against `master` in turn.

Let me know if you'd prefer to continue handling #507 and #509 as two separate PRs or if you want me to close #507 and we can review all the changes here.

martindemello:
yeah, i think it makes sense to merge the changes from #507 into this one and handle it as one big PR

pganssle:
Ok, I've closed #507. This PR already included all the changes in #507, so no changes are needed here.

merwok:
I am surprised that the setup worked at all before: `pytype/overlays` is not a valid package name, should have been `pytype.overlays` :open_mouth: 

Not relevant for this repo anymore, but you may have other projects with the same error.

PR #534, opened 2020-03-20
Google sync

rchen152:


rchen152:
Tests failing because of the use of assertCountEquals :( Will close and try again.

PR #542, opened 2020-03-31
Disable lint error not-an-iterable.

rchen152:
The error is being reported here:
https://travis-ci.org/github/google/pytype/jobs/669392038. Looks like a
false positive to me.


rchen152:
ping?

PR #554, opened 2020-04-07
Call sys.exit() with the return code in build_scripts/pyexe.

rchen152:
With this fix, our Travis script will report type-checking failures.


rchen152:
ping?

PR #556, opened 2020-04-09
Skips built in functions while merging pyi.

### LINKED ISSUE #550: merge-pyi feature request: do not annotate builtin functions
After using merge-pyi, the code base is littered with new annotations to `__str__`, `__eq__`, `__hash__` etc., which I don't bother annotating because they have well known data types. Would it be possible to not annotate these methods, when they match the expected builtin types?

souravbadami:
Fixes #550.

### DIFF
--- a/pytype/tools/merge_pyi/merge_pyi.py
+++ b/pytype/tools/merge_pyi/merge_pyi.py
@@ -626,10 +626,11 @@ class FixMergePyi(BaseFix):
       return []
 
   def can_annotate(self, src_sig):
-    if src_sig.short_name.startswith('__') and \
-            src_sig.short_name.endswith('__') and \
-            src_sig.short_name != "__init__":
-      self.logger.warning('builtin function, skipping %s', src_sig)
+    if (src_sig.short_name.startswith('__') and
+        src_sig.short_name.endswith('__') and
+        src_sig.short_name != "__init__" and
+        src_sig.short_name != "__new__"):
+      self.logger.info('magic method, skipping %s', src_sig)
       return False
 
     if src_sig.has_pep484_annotations or src_sig.has_comment_annotations:
--- a/pytype/tools/merge_pyi/merge_pyi.py
+++ b/pytype/tools/merge_pyi/merge_pyi.py
@@ -627,7 +627,8 @@ class FixMergePyi(BaseFix):
 
   def can_annotate(self, src_sig):
     if src_sig.short_name.startswith('__') and \
-            src_sig.short_name.endswith('__'):
+            src_sig.short_name.endswith('__') and \
+            src_sig.short_name != "__init__":
       self.logger.warning('builtin function, skipping %s', src_sig)
       return False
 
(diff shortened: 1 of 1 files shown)


souravbadami:
@rchen152 - Could you please review the changes?

rchen152:
Will review today, thanks for the PR!

pytype/tools/merge_pyi/merge_pyi.py
```diff
@@ -626,6 +626,12 @@ def insert_comment_annotation(self, src_sig, pyi_sig):
       return []
 
   def can_annotate(self, src_sig):
+    if src_sig.short_name.startswith('__') and \
```
rchen152, line 629: We follow the Google Python style guide, which forbids backslash line continuations (http://google.github.io/styleguide/pyguide.html#32-line-length). Could you find another way to break up this expression?

pytype/tools/merge_pyi/merge_pyi.py
```diff
@@ -626,6 +626,12 @@ def insert_comment_annotation(self, src_sig, pyi_sig):
...
+            src_sig.short_name.endswith('__') and \
+            src_sig.short_name != "__init__":
```
rchen152, line 631: In addition to `__init__`, please also check for `__new__`.

pytype/tools/merge_pyi/merge_pyi.py
```diff
@@ -626,6 +626,12 @@ def insert_comment_annotation(self, src_sig, pyi_sig):
...
+      self.logger.warning('builtin function, skipping %s', src_sig)
```
rchen152, line 632: nit: I find "builtin function" slightly confusing, since it makes me of think of builtins (https://docs.python.org/3/library/builtins.html). "Magic method" is a more common way of referring to these.

kamahen, line 632: Shouldn't this be info? It'll be happening a lot and it's not an indication of anything going wrong (a warning is indication that something is probably OK but should be checked)

pytype/tools/merge_pyi/merge_pyi.py
souravbadami: Yeah, used parentheses for an implicit line joining.

pytype/tools/merge_pyi/merge_pyi.py
souravbadami: Done. But I'm not much familiar with this codebase. Can you give me a reference to read more about this change?

pytype/tools/merge_pyi/merge_pyi.py
souravbadami: Yeah, done.

souravbadami: Oh, exactly. My bad. Done.

souravbadami:
@rchen152 - Resolved. Please review.

pytype/tools/merge_pyi/merge_pyi.py
rchen152: Do you mean a reference for `__new__`? It runs before `__init__` to create an instance of a class: https://docs.python.org/3/reference/datamodel.html#object.__new__.

souravbadami: Oh, that way. Got it. Thanks!

PR #558, opened 2020-04-09
Test against 3.8 and master with allowed failures.

pganssle:
This adds non-required jobs that will test against 3.8 and the master branch, which should make it easier to determine what needs to be done to add support for those Python versions.

Note: This will currently always fail trivially, because it's excluded in `python_requires`. This will still be beneficial because non-maintainers can modify `setup.cfg` in their PR branches, but they cannot modify `.travis.yml`. A better long-term solution would be to add a workaround that allows installing the package on nominally unsupported versions of Python

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,14 @@ language: python
 
 matrix:
   include:
-  - python: "3.5"
-  - python: "3.6"
-  - python: "3.7"
+    - python: "3.5"
+    - python: "3.6"
+    - python: "3.7"
+    - python: "3.8"
+    - python: "nightly"
+  allow_failures:
+    - python: "3.8"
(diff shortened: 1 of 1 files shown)


pganssle:
Oh hm, I guess Travis is configured to let me modify it even in a PR branch, since apparently my changes took effect immediately.

pganssle:
Also I forgot about #376 - this doesn't even pip install anything, so it apparently will actually run the tests against 3.8 and master without any sort of hack or workaround (well, arguably the way the tests are run now might count as a "hack", but at least no *additional* hacks...).

martindemello:
this is subsumed by PR #559, right? might as well close this one and move all the discussion there.

Issue #440, opened 2019-10-23
Support Python 3.8


cjolowicz:
Invoking pytype under Python 3.8 fails with the following error:

```
Python versions > 3.7 are not yet supported.
```

Is there an ETA for Python 3.8 support?

kamahen:
https://bugs.python.org/issue36541 contains a proposed PR to bring lib2to3 up-to-date. But I think pytype also uses another parser?

staticdev:
Any expectations on this one? Starting a new 3.8 project, and deciding if using pytype may be viable in some timeframe.

rchen152:
No progress to report yet, unfortunately. We've run into some unexpected difficulties trying to get internal testing set up for minor version upgrades.

ibozhilov:
I saw that you are working on setting up testing for python 3.8 https://github.com/google/pytype/pull/559. Do you have any update on how much work it will take to support python 3.8.

gpshead:
The opcode changes in Python 3.8:
  https://github.com/python/cpython/commit/520b7ae27e39d1c77ea74ccd1b184d7cb43f9dcb

staticdev:
I am pretty sure @rchen152 make a very nice job at pytype, but this project is just too big for one person. Unfortunately, I had to drop pytype in new projects in favor to mypy and pyre that are more maintained and support latest python. We use type checkers to ensure quality, and projects normally can't wait this long.

gpshead:
Use whatever works for you.  There are several of us interally actively working on pytype where it is used daily on tens of millions of lines of Python code.  We obviously haven't focused a lot on externally usability ease like this as other projects such as mypy have.  pytype's type inferencing is valuable, but only when you can _actually run it_ in your environment.  :)

rchen152:
Apologies for the lack of response. We discussed this at our last team meeting, and @martindemello and @Solumin have gamely volunteered to take a stab at implementing basic 3.8 support when their schedules allow. (New features, such as `typing.TypedDict`, will take longer.)

The problem is that, internally at Google, the need for 3.8 support is still ~months out, which makes it hard to prioritize above other features. We do hope to naturally get better at minor version upgrades as we do more of them, but we're also exploring whether it's feasible to deliver support for new versions externally on an faster timeline - no promises yet, though.

kamahen:
If you need lib2to3-like parser support, I've written up a proposal to replace lib2to3 with an augmented stdlib ast module:
[...]

PR #559, opened 2020-04-09
Initial stab at Python 3.8 support

pganssle:
The unusual testing setup makes this very hard to test locally, but this PR is an attempt to get an idea of what exactly needs to be done to support Python 3.8+ in pytype.

### DIFF
--- a/pytype/typegraph/cfg.cc
+++ b/pytype/typegraph/cfg.cc
@@ -504,7 +504,11 @@ PyTypeObject PyProgram = {
   tp_basicsize : sizeof(PyProgramObj),
   tp_itemsize : 0,
   tp_dealloc : ProgramDealloc,
+#if PY_VERSION_HEX >= 0x03080000 // 3.8
+  tp_vectorcall_offset: -1,
+# else
   tp_print : nullptr,
+# endif
   tp_getattr : nullptr,
   tp_setattr : nullptr,
 #if PY_VERSION_HEX >= 0x03050000  // 3.5
@@ -773,7 +777,11 @@ PyTypeObject PyCFGNode = {
   tp_basicsize : sizeof(PyCFGNodeObj),
   tp_itemsize : 0,
   tp_dealloc : CFGNodeDealloc,
+#if PY_VERSION_HEX >= 0x03080000 // 3.8
+  tp_vectorcall_offset: -1,
+# else
   tp_print : nullptr,
+# endif
   tp_getattr : nullptr,
   tp_setattr : nullptr,
 #if PY_VERSION_HEX >= 0x03050000  // 3.5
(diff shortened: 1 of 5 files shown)


martindemello:
@pganssle thanks for the PR! i'd also like to know more about the problems you had with running the tests; could you open a bug for them?

martindemello:
oh, i see from the comments in the other PR that you are talking about the `tox` and src layout issue, right?

pytype/typegraph/cfg.cc
```diff
@@ -504,7 +504,11 @@ PyTypeObject PyProgram = {
   tp_basicsize : sizeof(PyProgramObj),
   tp_itemsize : 0,
   tp_dealloc : ProgramDealloc,
+#if PY_VERSION_HEX >= 0x03080000 // 3.8
```
rchen152, line 507: @Solumin FYI

This change looks harmless enough to me but I'm not familiar with this code.

Solumin: Yeah, this all looks good: https://www.python.org/dev/peps/pep-0590/#changes-to-the-pytypeobject-struct

rchen152: Thanks!

pganssle: Ah yeah, FYI I explained each change in the corresponding commit message, see: https://github.com/google/pytype/pull/559/commits/cde5e9c4b0413d8123d64c0d4400efe91e8e64e5

Though I realize that's probably not the best way to communicate it in a PR...

rchen152: Ahhh I totally missed that ><

pganssle:
> @pganssle thanks for the PR! i'd also like to know more about the problems you had with running the tests; could you open a bug for them?

Yeah at some point I will open a bug for them, but it's not just `tox` and `src/` environment. These tests require you to have Python 2.7 installed for some reason, they can't be run from a virtual environment, and the outputs are confusing and difficult to interpret.

For example, I don't know what to do next on this PR because the tests fail, but there's no output as to why they failed:

https://travis-ci.org/github/google/pytype/jobs/673128609

When I run them locally (I needed to create a dedicated Python install for this because of the virtual environment problem), it gave me a list of failing tests, and told me to look at the corresponding logs files, all of which were empty.

If you have any tips on how to figure out why these are failing let me know. I was kinda hoping that the travis build would be giving useful information so I could unblock myself, but alas no.

rchen152:
Sorry for letting this sit for so long. Looking back through the comments, I don't think there are any issues that would prevent this from being merged (there's a nit about the placement of a lint directive), so I'm going to go ahead and merge it so the continuation in https://github.com/google/pytype/pull/587 is easier to review.

This PR touches both GitHub-first and google3-first files, so I'm going to merge it directly, then copy the relevant portion of the changes internally and do an export. The commit history will look a little wonky, but hopefully everything will end up in the right state in the end.

martindemello:
addresses #440

PR #563, opened 2020-04-16
Strays 'nothing' type for empty tuple inference.

### LINKED ISSUE #549: Stray 'nothing' type for empty tuple inference
While using `merge-pyi -i`, an unannotated function that returns an empty tuple got as type `Tuple[nothing, ...]`. Since this type is not defined, a later pytype run fails. If the function is already annotated, nothing happens.

### Input file:

```
from typing import Tuple

def empty_tuple1():
  return ()

def empty_tuple2() -> Tuple:
  return ()
```

### Output pyi
```
# (generated with --quick)

from typing import Tuple

def empty_tuple1() -> Tuple[nothing, ...]: ...
def empty_tuple2() -> tuple: ...
```

souravbadami:
Fixes #549.


pytype/pytd/pytd_visitors.py
```diff
@@ -694,7 +699,9 @@ def MaybeCapitalize(self, name):
   def VisitGenericType(self, node):
     """Convert a generic type to a string."""
     parameters = node.parameters
-    if self._NeedsTupleEllipsis(node):
+    if self._IsEmptyTuple(node):
+      return self.MaybeCapitalize(node.base_type) + "[()]"
```
martindemello, line 703: just set `parameters = "()"` here

martindemello: sorry, `parameters = ("()",)` since it needs to be a tuple

souravbadami: Oh, yeah. I could have done that. 
I need to read a bit more and get familiar with the codebase.
Resolved.

souravbadami:
@rchen152 - Could you please review the changes?

souravbadami:
I can't merge it as I don't have permission to do it. Thank you!

rchen152:
Heh, that's true.

FYI the export failed because of an unrelated change that broke some tests - hopefully I'll get it sorted out on Monday.

PR #569, opened 2020-04-23
Add a requirements.txt.

rchen152:
With this, we can run `pip install -r requirements.txt` to get all the
dependencies for building and testing the GitHub code locally.
Previously, I was looking at .travis.yml every time to figure out what
to install.

Based on
https://packaging.python.org/discussions/install-requires-vs-requirements/,
setup.py's install_requires and requirements.txt accomplish different
purposes, so I haven't tried to keep them in sync.

I'm also hoping that adding this file will fix our broken dependency
graph, which currently shows no dependencies and no dependents:
https://help.github.com/en/github/visualizing-repository-data-with-graphs/listing-the-packages-that-a-repository-depends-on.

I also deleted a stray newline in CONTRIBUTING.md, because why not?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,10 @@
+# Python dependencies for building and testing the pytype source code.
+# Make sure you also install the non-Python dependencies described in
+# https://github.com/google/pytype/blob/master/CONTRIBUTING.md#pytype-dependencies.
+# In particular, use the official ninja package, not the pip distribution.
+attrs
+importlab
+pylint
+pyyaml
+six
+typed_ast
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,6 @@ frustration later on.
 All submissions, including submissions by project members, require review. We
 use Github pull requests for this purpose.
 
-
 ### Pytype dependencies
 Before you can build and test Pytype, you will have to install a few
 dependencies.


rchen152:
ping?

requirements.txt
```diff
@@ -0,0 +1,10 @@
+# Python dependencies for building and testing the pytype source code.
+# Make sure you also install the non-Python dependencies described in
+# https://github.com/google/pytype/blob/master/CONTRIBUTING.md#pytype-dependencies.
+# In particular, use the official ninja package, not the pip distribution.
```
martindemello, line 4: if the pip installed ninja doesn't work, we should repeat this advice in CONTRIBUTING.md too.

rchen152: CONTRIBUTING.md links directly to ninja-build.org, which provides instructions on how to get the official package.

The only reason I mention the pip distribution here is that if someone checks setup.cfg, they might be confused by we have developers and users install ninja from different sources. Pip installing ninja won't cause any terrible problems, just make debugging harder in the unlikely event of the PyPI package (which isn't maintained by ninja, but by some ninja user that wanted ninja to be pip-installable for their own project) being broken or out-of-date.

rchen152:
I accidentally deleted this file with our last google3->GitHub export :woman_facepalming: I'm just going to add it back directly rather than creating another pull request, since the file has already gone through review.

PR #587, opened 2020-05-14
Add minimum changes required for py3.8 support

ammaraskar:
This is a continuation of @pganssle's work on #559 

Beyond just making it compile, this also adds some of the newer VM and marshaling stuff such as positional only arguments. It also addresses and adds tests for a 3.6 feature that allows the line number to opcode table (`lnotab`) to be non monotonic. After this the remaining failing tests are:

```
** pytype.vm_test
** pytype.tests.py3.test_variable_annotations
** pytype.tools.xref.indexer_test
** pytype.tests.py3.test_typing_namedtuple
** pytype.tests.test_decorators
** pytype.tests.test_with
** pytype.tests.test_exceptions
** pytype.tests.test_flow
** pytype.tests.py3.test_coroutine
** pytype.tests.py3.test_annotations
** pytype.tests.py3.test_stdlib
** pytype.tests.test_builtins1
** pytype.tests.test_basic
```

I'll take a look at them later or someone else can carry through with those. Either way I'd like to land this first to make the reviews manageable. 

-----

As @pganssle noted on their PR, there is some weird behavior going on with the tests that causes them to fail but not produce any output in the console or the log files.

Upon further investigation this was caused by a segmentation fault in `cfg.cc`. In Python 3.8, `PyStructSequence_InitType` was changed so that it errors out if a type with a reference count greater than 0 is passed in: https://github.com/python/cpython/blame/3.8/Objects/structseq.c#L396-L400

This error was being silently swallowed because we were just assuming that `PyStructSequence_InitType` was running successfully. The problem was we were using `PyObject_HEAD_INIT` which [initializes the refcount to 1](https://github.com/python/cpython/blob/35b8a4da7494cd51f3edeaa07764dd8f14bac0bf/Include/object.h#L83-L85).

With this fixed there should now be actual output from the test runs so hopefully this work should be easier to continue by others.

----

### Notes on the workflow

As Paul mentioned, the workflow on the open-source end of pytype is pretty bespoke and not too friendly. I would recommend using `tox` and separating out the `googletest` based C/C++ tests from the Python ones. The `build_scripts` folder has a lot of magic that masks the details of how the build system works underneath. CMake is currently being used to build the extension, run Python tests and run C++ tests. Building and testing C++ is CMake's forte but there's a lot of code to handle the discovery and running of Python tests. This might have been done to keep symmetry with the `py_test` build rule internally but it is completely alien to most open-source Python developers. Instead, using something like the builtin `unittest` module or `pytest` to drive the Python tests would be a more comfortable experience.

For example, one of the issues I had was figuring out how to make CMake use my locally compiled Python 3.8 build. Passing custom arguments to CMake is nearly impossible due to the build_scripts. I'd recommend this, if you're on a debian based distro, add the deadsnakes ppa or similar and use:
```shell
sudo apt install python3.7 python3.7-dev python3.7-venv
python3.7 -m venv venv
. ./venv/bin/activate
python3.7 build_scripts/build.py
python3.7 build_scripts/run_tests.py
```
Swap out 3.7 for whatever host Python version you want to test with.

Next, for debugging issues like the mysterious failing tests without output, the most important file is `out/ninja.log`. This actually shows the commands being invoked to run the tests. The exact incantation seems to be:

`TYPESHED_HOME=/home/aaskar/workspace/pytype/typeshed /home/aaskar/workspace/pytype/venv/bin/python3.8 -B /home/aaskar/workspace/pytype/build_scripts/test_module.py pytype.tests.test_basic -o /home/aaskar/workspace/pytype/out/pytype/tests/test_basic.log -P /home/aaskar/workspace/pytype/out -s -p`

Failing tests can also be run with unittest instead of the custom runner for a workflow where you don't have to repeatedly check a log file after making changes like so:

`TYPESHED_HOME=$PWD/typeshed python3.8 -m unittest pytype.tools.traces.traces_test`


ammaraskar:
Thanks for merging the parent PR :)

Rebased and looking into the test failures.

rchen152:
Thanks for resolving the test failures! @martindemello FYI this is probably ready for review.

pytype/typegraph/cfg.cc
```diff
@@ -812,11 +812,7 @@ PyTypeObject PyCFGNode = {
 
 // --- Origin --------------------------------------------------------------
 
-static PyTypeObject PyOrigin = {
-  // initialized in inittypegraph
-  PyVarObject_HEAD_INIT(nullptr, 0)
-  tp_name: nullptr
-};
+static PyTypeObject PyOrigin;
```
martindemello, line 815: why this change?

pytype/vm.py
```diff
@@ -2507,6 +2507,20 @@ def byte_SETUP_FINALLY(self, state, op):
     self.store_jump(op.target, state.push(self.convert.build_none(state.node)))
     return self.push_block(state, "finally", op, op.target)
 
+  # New python3.8+ exception handling opcodes:
+  # BEGIN_FINALLY, END_ASYNC_FOR, CALL_FINALLY, POP_FINALLY
+  def byte_BEGIN_FINALLY(self, state, op):
+    return state
```
martindemello, line 2513: we need to handle the stack effects for all these opcodes, and also check for the `NULL` pushed by `BEGIN_FINALLY` in `WITH_CLEANUP_START` and `WITH_CLEANUP_FINISH`

pytype/typegraph/cfg.cc
ammaraskar: This is the problem explained in my initial comment:

> This was caused by a segmentation fault in cfg.cc. In Python 3.8, PyStructSequence_InitType was changed so that it errors out if a type with a reference count greater than 0 is passed in: https://github.com/python/cpython/blame/3.8/Objects/structseq.c#L396-L400
> 
> This error was being silently swallowed because we were just assuming that PyStructSequence_InitType was running successfully. The problem was we were using PyObject_HEAD_INIT which initializes the refcount to 1.

The recommended way to use PyStructSequence is this, as you can see in the stdlib:

* https://github.com/python/cpython/blob/578c3955e0222ec7b3146197467fbb0fcfae12fe/Objects/longobject.c#L5597
* https://github.com/python/cpython/blob/a487a39dca4c41305928c7dfdbcb0b3aa344683b/Python/thread.c#L164

pytype/vm.py
ammaraskar: Let me add some `TODO`s here, handling the vm is a bit complicated here since 3.8 kinda revamped how error handling is done. Doing both the changes makes the diff really large, so I'd rather get this merged in to get the basic tests working and then tackle this separately.

ammaraskar: For reference: https://github.com/python/cpython/commit/520b7ae27e39d1c77ea74ccd1b184d7cb43f9dcb

martindemello: okay, sounds good

pytype/tools/traces/traces_test.py
```diff
@@ -283,28 +283,40 @@ def f(x, y, z):
...
+    matches = self._get_traces(
+        "v = 42",
+        ast.Constant if sys.version_info >= (3, 8) else ast.Num)
     self.assertTracesEqual(matches, [((1, 4), "LOAD_CONST", 42, ("int",))])
 
   def test_str(self):
-    matches = self._get_traces("v = 'hello'", ast.Str)
+    matches = self._get_traces(
+        "v = 'hello'",
+        ast.Constant if sys.version_info >= (3, 8) else ast.Str)
```
martindemello, line 294: break all these into two tests, e.g. `test_str` and `test_str_38`, and guard the tests with `@testutils.skipBeforePy`/`@testutils.skipAfterPy` (see `test_literal_37` above for an example).

pytype/pyc/opcodes.py
```diff
@@ -319,6 +319,15 @@ class BEFORE_ASYNC_WITH(Opcode):
   __slots__ = ()
 
 
+class BEGIN_FINALLY(Opcode):
+  __slots__ = ()
+
+
+class END_ASYNC_FOR(Opcode):
+  FLAGS = HAS_JUNKNOWN | NO_NEXT # might re-raise an exception
```
martindemello, line 327: lint: two space before the `#`

pytype/pyc/opcodes.py
```diff
@@ -835,6 +844,17 @@ class CALL_METHOD(OpcodeWithArg):  # Arg: #args
   __slots__ = ()
 
 
+class CALL_FINALLY(OpcodeWithArg):  # Arg: Jump offset to finally block
+  FLAGS = HAS_JREL|HAS_ARGUMENT
```
martindemello, line 848: lint: add spaces around the `|`

pytype/pyc/opcodes.py
```diff
@@ -835,6 +844,17 @@ class CALL_METHOD(OpcodeWithArg):  # Arg: #args
...
+  __slots__ = ()
+
+
+class POP_FINALLY(OpcodeWithArg):
+  FLAGS = HAS_ARGUMENT|HAS_JUNKNOWN  # might re-raise an exception or
```
martindemello, line 853: move the comment to above the line of code if it doesn't fit on one line.

pytype/pyc/opcodes_test.py
```diff
@@ -191,5 +200,44 @@ def test_call_method(self):
...
+    # Compiled from:
+    # f(
+    #   1,
+    #   2
+    #  )
+    code = [
+        0x65, 0,   # LOAD_NAME, arg=0th name
+        0x64, 0,   # LOAD_CONST, arg=0th constant
+        0x64, 1,   # LOAD_CONST, arg=1st constant
+        0x83, 0x2, # CALL_FUNCTION, arg=2 function arguments
```
martindemello, line 223: lint: needs at least two spaces before the `#`

pytype/pyc/opcodes_test.py
```diff
@@ -191,5 +200,44 @@ def test_call_method(self):
...
+    expected = [
+        ('LOAD_NAME', 0),
+        ('LOAD_CONST', 0),
+        ('LOAD_CONST', 1),
+        ('CALL_FUNCTION', 2),
+        ('RETURN_VALUE',)
+    ]
+    self.assertDisassembly(code, expected)
+    lnotab = [
+        0x2, 0x1, # +2 addr, +1 line number
```
martindemello, line 235: two spaces here too

pytype/pyc/opcodes_test.py
```diff
@@ -28,6 +29,14 @@ def assertDisassembly(self, code, expected):
       else:
         self.assertEqual(e, (o.name, o.arg))
 
+  def assertLineNumbers(self, code, co_lnotab, expected):
+    """Assert that the opcodes in a code sequence have the expected
```
martindemello, line 33: see if you can fit this on one line; if not, the linter requires multiline doc strings to have complete documentation of args. also needs a `.` at the end of the first line.

martindemello:
thanks, this looks good, apart from the travis failure. i'll test it internally and merge it

ammaraskar:
The travis failure looks new and unrelated, getting it on a clean build without this as well:

```
FAIL: pytype.tests.test_import.ImportTestPy2.test_imported_method_as_class_attribute
Traceback (most recent call last):
  File "/home/aaskar/workspace/pytype_staging/out/pytype/tests/test_import.py", line 565, in test_imported_method_as_class_attribute
    """)
  File "/home/aaskar/workspace/pytype_staging/out/pytype/tests/test_base.py", line 427, in assertTypesMatchPytd
    self.assertMultiLineEqual(pytd_tree_src, ty_src)
AssertionError: 'os: [17 chars]oo:\n    def killpg(__pgid: int, __signal: int) -> None: ...\n' != 'os: [17 chars]oo:\n    def killpg(pgid: int, sig: int) -> None: ...\n'
  os: module
  
  class Foo:
-     def killpg(__pgid: int, __signal: int) -> None: ...
?                --           --   ---
+     def killpg(pgid: int, sig: int) -> None: ...
```

rchen152:
test_imports was changed recently when I updated the typeshed commit that pytype is pinned to (https://github.com/google/pytype/commit/2e820695104dd5e93d2297fd12c1e26a5c43b3df). My guess is that the test and/or the typeshed submodule is out-of-date in your fork.

ammaraskar:
Aah thanks, I had retained the old `typeshed` submodule locally. Weird that travis is doing the same thing.

ammaraskar:
Yeah not sure what's up with travis, it's pulling out `typeshed@d5851eca6f448b648e1a7b3f3010cc3dde498535`
https://travis-ci.org/github/google/pytype/jobs/691545021#L348

despite the fact that it's pinned to:
`5fe6a5b`

Gonna try re-opening and closing.

ammaraskar:
There we go, Travis is back to passing :)

martindemello:
addresses #440

martindemello:
all tests passed; i'll merge this internally and close the pull request (you will still be listed as the commit author when we reexport the changes)

Issue #562, opened 2020-04-16
set up a build system for pytype binary wheels


rchen152:
since pytype's source installation depends on a c compiler, it would be nice to have binary wheels uploaded to pypi. this requires building the wheel on the target machine. we need each of
* manylinux (docker images available)
* windows (need a vm with a c compiler and toolchain)
* mac (???)

IaroslavR:
It can be done with the help of https://github.com/pypa/auditwheel I done some researches in https://github.com/IaroslavR/aws-lambda-builder and except of one problem - https://github.com/pypa/auditwheel/issues/159 it's a pretty straightforward  process

bsolomon1124:
I am interested in helping out with this and have some experience building platform wheels in other projects.  GitHub Actions, Travis, and Azure DO all have support for Windows and MacOSX, I believe.

bsolomon1124:
Some notes from Mac OS X 10.15.4 below.

Sequence of commands used to build macosx wheel:

```
$ git clone --recurse-submodules https://github.com/google/pytype.git
$ cd pytype
$ pyenv local 3.8.3
$ python -m venv venv
$ . ./venv/bin/activate
$ python -m pip install -U pip wheel setuptools
$ CMAKE_PREFIX_PATH="$(brew --prefix bison)" python setup.py bdist_wheel
$ ls dist/*.whl
dist/pytype-2020.5.13-cp38-cp38-macosx_10_15_x86_64.whl
```

Notes on the dependencies from https://github.com/google/pytype/blob/master/CONTRIBUTING.md#pytype-dependencies:

**A C++11 compiler for your platform**: Mac OS X masks `g++` to `clang`.  (It's not a symlink; I'm actually not sure what it is, but...)

```
$ g++ --help | head -n3
OVERVIEW: clang LLVM compiler

USAGE: clang [options] file...
```

Also, I'm not sure if `/usr/bin/g++` exists out of the box or needs `xcode-select --install` to be put there.

**CMake version 2.8 or higher**:

This one is easy, `brew install -v cmake`.

**Bison version 3.0.2 or higher**: 

This one is a bit tricky because in my case I had `/usr/bin/bison` version 2.3.  (Again, not sure if this was from `xcode` or there already.)

So after using `brew install -v bison`, which is "keg-only", I needed the `CMAKE_PREFIX_PATH` to tell CMake where to find Bison version 3.x.

[...]

PR #593, opened 2020-05-20
Wheels

bsolomon1124:


bsolomon1124:
Well, I was mistaken on this one.  It looks like Travis dropped support for Python on MacOS over a year ago:

- https://docs.travis-ci.com/user/languages/python/
- https://github.com/travis-ci/travis-ci/issues/9929

There may be a workaround with using something like pyenv or (reportedly) an older version of xcode, but not sure.

bsolomon1124:
Okay, so it looks like we have successful builds for Python 3.5 thru 3.7 on MacOS:

https://travis-ci.org/github/google/pytype/builds/689428477

Some food for thought:

- Because Travis does not officially support `language: python` on MacOS, this uses `language:c++` and `pyenv` as a workaround. Very much up to the pytype developers whether they feel that is stable and acceptable way to do things. Otherwise, building MacOS wheels would probably require a switch to different CI vendor.
- Question from me (I'm not well versed with ninja, cmake, or the pytype build ecosystem): what is the difference between building a wheel the standard way and with whatever `build_scripts/build.py` does?
- The Travis stdout shows a warning about `MACOSX_RPATH`.  Not sure if that deserves any attention.

Will move onto [`manylinux`](https://github.com/pypa/manylinux) wheels next.

rchen152:
> Question from me (I'm not well versed with ninja, cmake, or the pytype build ecosystem): what is the difference between building a wheel the standard way and with whatever build_scripts/build.py does?

pytype's build system is similar to [bazel](https://en.wikipedia.org/wiki/Bazel_(software)) but implemented using cmake + ninja because they're lighter-weight and more common dependencies. So the"build" step compiles the C++ extensions, etc., and puts some runnable binaries in pytype/out.

We did this because it's familiar to us, but it's become increasingly obvious that this is confusing to contributors - we're planning to add more docs and hopefully make some changes to make the build system less non-standard in the next couple of weeks.

bsolomon1124:
Okay, so I've made it part way with manylinux wheels (actually looks promising, I think) but am getting stuck on a `cmake` issue with a call to [`FindPythonLibs()`](https://cmake.org/cmake/help/v3.5/module/FindPythonLibs.html).

In this case I'm working in the manylinux2010 container:

```
docker container run -it --rm --volume "$(pwd)":/io quay.io/pypa/manylinux2014_x86_64
```

With this build script:

> https://github.com/google/pytype/blob/590058a161a6a705fef979059a817009c9c113cb/build_scripts/manylinux.sh

What I can't seem to get around is the FindPythonLibs call:

bsolomon1124:
Looks like using a more recent version of CMake [may address this](https://github.com/pypa/manylinux/issues/613), confirming that now.

bsolomon1124:
Okay, looks like we have [manylinux2014 wheels](https://travis-ci.org/github/google/pytype/jobs/690678227) building green also.

One thing I'm unclear on at this point is regarding the `libpythonX.Y.so` library that `cmake`'s FindPythonLibs looks for.  By default on the image you have:

```
$ ls /usr/lib64/libpython*
/usr/lib64/libpython2.7.so.1.0
```

Then installing `python3-devel` looks to be for 3.6:

```
$ yum install -y -q python3-devel
$ ls -1 /usr/lib64/libpython*
/usr/lib64/libpython2.7.so.1.0
/usr/lib64/libpython3.6m.so
/usr/lib64/libpython3.6m.so.1.0
/usr/lib64/libpython3.so
```

Even when the build process is invoked with e.g. Python 3.7, CMake seems content using `/usr/lib64/libpython3.6m.so`.  I'm unsure (from not knowing CPython nearly well enough at that level of things) whether that's okay, or whether not using a `libpython3.7m.so` threatens to break anything.  It looks like yum provides a `python34-devel`, `python36-devel`, but not 35 or 37.

bsolomon1124:
Also, confirming that these wheels should not requiring any "bundling" of third-party shared libraries (as done by `auditwheel` and `delocate`) that are outside of the [manylinux2014 whitelist](https://www.python.org/dev/peps/pep-0599/#the-manylinux2014-policy):

```
$ objdump -x pyi/parser_ext.cpython-37m-x86_64-linux-gnu.so | grep NEEDED
  NEEDED               libstdc++.so.6
  NEEDED               libm.so.6
  NEEDED               libgcc_s.so.1
  NEEDED               libpthread.so.0
  NEEDED               libc.so.6
```

All of those are from the whitelist.

And, on another positive note, what is _not_ present there is `libpython`.  That is a good thing because the manylinux policy [advises against this](https://www.python.org/dev/peps/pep-0513/#libpythonx-y-so-1).  `libpython` in this case is needed by cmake but not a dependency of the extension module itself.

bsolomon1124:
One thing missing at this juncture is the artifacts themselves. Unless I'm missing something, I believe Travis CI wants [needs AWS credentials](https://docs.travis-ci.com/user/uploading-artifacts/) to put them anywhere.

The alternatives to Travis CI include GitHub Actions and Azure DevOps.  (Mentioning these two specifically because I am fairly sure they have macOS support.)  Though, if I recall, the Python SDK on GitHub Actions macOS is 10.13 - you are pegged at that version rather than being able to use a matrix.

pganssle:
> One thing missing at this juncture is the artifacts themselves. Unless I'm missing something, I believe Travis CI wants [needs AWS credentials](https://docs.travis-ci.com/user/uploading-artifacts/) to put them anywhere.
> 
> The alternatives to Travis CI include GitHub Actions and Azure DevOps. (Mentioning these two specifically because I am fairly sure they have macOS support.) Though, if I recall, the Python SDK on GitHub Actions macOS is 10.13 - you are pegged at that version rather than being able to use a matrix.

If you need a template for Github actions, I've recently added automatic build and PyPI release to [the zoneinfo backport](https://github.com/pganssle/zoneinfo/blob/master/.github/workflows/build-publish.yml). To the extent that that's not just standard boilerplate, I wrote it all myself, so it's Google-owned.

I personally have GHA do the PyPI publishing and it's pretty dead-simple, but I dunno if pytype has some copybara-based wheel publisher or something.

build_scripts/wheels/macos.sh
```diff
@@ -0,0 +1,19 @@
...
+pyenv --version
+pyenv install -s "$PYENV_VERSION"
+pyenv local "$PYENV_VERSION"
+python -V
+brew install -v bison || brew upgrade bison
+brew install -v cmake || brew upgrade cmake
+brew install -v ninja || brew upgrade ninja
+python -m pip install -U wheel setuptools
+export CMAKE_PREFIX_PATH="$(brew --prefix bison)"
+python setup.py bdist_wheel -d macos-wheelhouse
```
pganssle, line 19: I recommend replacing this with a PEP 517 build front-end.

Both `pip` and `pep517` will work (I tend to use `pep517.build` because it's at least designed for this, though it was somewhat grudgingly added).

`pip` version is:

```suggestion
export CMAKE_PREFIX_PATH="$(brew --prefix bison)"
python -m pip wheel . --no-deps -w macos-wheelhouse
```

The `pep517` version is:

```suggestion
python -m pip install -U pep517
export CMAKE_PREFIX_PATH="$(brew --prefix bison)"
python -m pep517.build -b -o macos-wheelhouse
```

Same comment applies in `manylinux.sh`.

bsolomon1124: I had actually been trying to use `pip wheel` at first, but it was hanging for a very long time, I think because I had forgotten the `--no-deps` flag.  Thanks for catching that, will make the change.

bsolomon1124:
FYI that quay.io is down as of this morning.

bsolomon1124:
Okay, looks like manylinux2014 + MacOS wheels are building successfully with GitHub Actions:

> https://github.com/bsolomon1124/pytype/actions

Thanks for the hand @pganssle.

Resulting wheels:

```bash
$ ls -1 ~/Downloads/dist/ | sort
pytype-2020.5.13-cp35-cp35m-macosx_10_14_x86_64.whl
pytype-2020.5.13-cp35-cp35m-manylinux2014_x86_64.whl
pytype-2020.5.13-cp36-cp36m-macosx_10_14_x86_64.whl
pytype-2020.5.13-cp36-cp36m-manylinux2014_x86_64.whl
pytype-2020.5.13-cp37-cp37m-macosx_10_14_x86_64.whl
pytype-2020.5.13-cp37-cp37m-manylinux2014_x86_64.whl
```

**Updated TODOs**:

- _Other `manylinux` images_: I chose `manylinux2014_x86_64` because the underlying CentOS is recent enough to support fairly easy install of the build ecosystem tools.  However, [manylinux2014](https://github.com/pypa/manylinux) is only recognized by `pip >= 19.3`.  Additionally, there are i686 images, though I did not experiment with that arch variant and don't know if it's feasible.

- _Update cadence for build dependencies_: `manylinux.sh` pins to specific versions of cmake/flex/bison for reproducibility, but might be a good idea to establish some type of patch cadence to receive security updates/bug fixes.

- _sdist + PyPI_: I neglected to also build an `sdist` in the GitHub workflow, or to upload to PyPI, both as shown in https://github.com/pganssle/zoneinfo/blob/master/.github/workflows/build-publish.yml.  Up to the maintainers of pytype how much of a human in the loop they would like to have in the upload process.

- _Trigger_: the `on` tag in the GitHUb workflow is currently just set to `push`, but [supports some other values that might offer more fine-grained control over when wheels are built](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#on)

- _Badge_: I added a badge that lights up green if wheels are detected on the PyPI pytype index.  This is obviously red at the moment.

rchen152:
Neat! I like the wheel badge :)

We tag all releases, so it would probably make sense to build a wheel every time a tag is pushed? Automatically uploading to TestPyPI could also be useful, but I think we'd want a human to test out the release before uploading it to PyPI proper.

rchen152:
Sorry for the delay in merging; I hadn't noticed that you removed the WIP label. Thank you for the contribution!

martindemello:
fixes #562

PR #605, opened 2020-06-20
Autodetect number of jobs with --jobs auto

lgeiger:
This adds the possibility to autodetect the number of jobs based on the number of available CPU cores on the host system using `pytype --jobs auto`. This allows for a usage similar to [`pytest-xdist`](https://github.com/pytest-dev/pytest-xdist#speed-up-test-runs-by-sending-tests-to-multiple-cpus) which can be useful when using `pytype` in scripts that might run on different systems.

### DIFF
--- a/pytype/tools/analyze_project/parse_args.py
+++ b/pytype/tools/analyze_project/parse_args.py
@@ -10,6 +10,7 @@ from pytype.tools.analyze_project import config
 
 _ARG_PREFIX = '--'
 
+
 def _auto_detect_cpus():
   try:
     return len(os.sched_getaffinity(0))
@@ -18,12 +19,15 @@ def _auto_detect_cpus():
 
 
 def parse_jobs(s):
-  """Parse the --jobs option"""
+  """Parse the --jobs option."""
   if s == 'auto':
     n = _auto_detect_cpus()
     return n if n else 1
   elif s is not None:
     return int(s)
+  else:
+    return None
+
 
 class Parser(arg_parser.Parser):
   """Subclasses Parser to add a config file processor."""
--- a/pytype/tools/analyze_project/parse_args.py
+++ b/pytype/tools/analyze_project/parse_args.py
@@ -1,6 +1,7 @@
 """Argument parsing for analyze_project."""
 
 import argparse
+import os
 
 from pytype import config as pytype_config
 from pytype.tools import arg_parser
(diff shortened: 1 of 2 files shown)


pytype/tools/analyze_project/parse_args.py
```diff
@@ -9,6 +10,20 @@
 
 _ARG_PREFIX = '--'
 
+def _auto_detect_cpus():
+  try:
+    return len(os.sched_getaffinity(0))
```
lgeiger, line 15: This return the number of CPUs the current process is restricted to on Linux.

kamahen: What's wrong with os.cpu_count()?

lgeiger: On Linux `len(os.sched_getaffinity(0))` returns the number of CPUs available to the current process, where as `os.cpu_count()` will return the total number of CPUs on the host system.

kamahen: answering my question - it's the number of available cores (not clear whether this includes hyperthreading cores).
More details: https://stackoverflow.com/questions/1006289/how-to-find-out-the-number-of-cpus-using-python/14840102

lgeiger: > not clear whether this includes hyperthreading cores

As far as I know this includes hyperthreading cores. On my dual-core macbook `--jobs auto` will result in 4 parallel jobs.

pytype/tools/analyze_project/parse_args.py
```diff
@@ -9,6 +10,20 @@
 
 _ARG_PREFIX = '--'
 
+def _auto_detect_cpus():
```
rchen152, line 13: Could you add a second newline before this function, and before the class on line 28? (We put two blank lines between top-level functions and classes.)

pytype/tools/analyze_project/parse_args.py
```diff
@@ -9,6 +10,20 @@
...
+  try:
+    return len(os.sched_getaffinity(0))
+  except AttributeError:
+    return os.cpu_count()
+
+
+def parse_jobs(s):
+  """Parse the --jobs option"""
```
rchen152, line 21: Please add a period at the end of the docstring.

pytype/tools/analyze_project/parse_args.py
```diff
@@ -9,6 +10,20 @@
...
+  if s == 'auto':
+    n = _auto_detect_cpus()
+    return n if n else 1
+  elif s is not None:
+    return int(s)
```
rchen152, line 26: Could you add a "return None" at the end of this method to make it obvious that it intentionally defaults to returning None?

lgeiger:
> Just a few nits related to the Google style guide.

Fixed in 4e48291

PR #678, opened 2020-09-18
Declare Python 3.8 support (whoo!) and add 3.9 to the testing matrix.

rchen152:
Python 3.9 is "3.9-dev" on Travis:
https://travis-ci.community/t/python-3-9-support/6703.

The m flag for manylinux Python tags is dropped in 3.8:
https://github.com/pypa/manylinux/blob/master/README.rst#manylinux2014.

Since it was trivial to do, I also silenced an incorrect lint error that
shows up in 3.9:
  "Value 'Union' is unsubscriptable (unsubscriptable-object)"
(or sometimes 'Optional').

For https://github.com/google/pytype/issues/440.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
           docker image pull "$DOCKER_IMAGE"
       - name: Build manylinux wheels
         env:
-          PYTHON_TAGS: 'cp35-cp35m cp36-cp36m cp37-cp37m'
+          PYTHON_TAGS: 'cp35-cp35m cp36-cp36m cp37-cp37m cp38-cp38'
         run: |
           docker container run \
             --rm \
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           docker image pull "$DOCKER_IMAGE"
       - name: Build manylinux wheels
         env:
-          PYTHON_TAGS: 'cp35-cp35m cp36-cp36m cp37-cp37m'
+          PYTHON_TAGS: 'cp35-cp35m cp36-cp36m cp37-cp37m cp38-cp38'
```
martindemello, line 30: should that be `cp38-cp38m`?

rchen152: The 'm' is gone in 3.8 (https://github.com/pypa/manylinux/blob/master/README.rst#manylinux2014): "Note that starting with CPython 3.8, default sys.abiflags became an empty string: the m flag for pymalloc became useless (builds with and without pymalloc are ABI compatible) and so has been removed. (e.g. /opt/python/cp38-cp38)"

martindemello (approved):

PR #693, opened 2020-10-01
Use list comprehensions for flattening

lgeiger:
This PR changes the visitor and typegraph utils to use list comprehensions for flattening nested sequences instead of `sum([...], [])` which can be slow for large inputs. This should improve readability and also slightly improve performance.


rchen152 (approved):
Thanks!

PR #695, opened 2020-10-02
Mention requirements.txt in CONTRUBUTING guide

josalmi:
It comes obvious when building pytype that these are required but was never mentioned in the CONTRIBUTING guide.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,6 +44,14 @@ dependencies.
    Hence, to run these tests, you will need Python 2.7 and Python 3.x
    (preferably 3.6+) interpreters installed on your system.
 
+Required Python packages are listed in the [requirements.txt](requirements.txt)
+file in this repository. They can be installed with pip with the following
+command:
+
+```shell
+pip install -r requirements.txt
+```
+
 The Pytype Git repository also contains few Git submodules. Before building
 the `pytype` executable or running tests, one has to ensure that the submodules
 are up to date. This can be done with the following command:


CONTRIBUTING.md
```diff
@@ -44,6 +44,13 @@ dependencies.
    Hence, to run these tests, you will need Python 2.7 and Python 3.x
    (preferably 3.6+) interpreters installed on your system.
 
+Required Python packages are listed in [requirements.txt](requirements.txt) file
+in this repository. This can be done with pip with the following command:
+
+```
+$> pip install -r requirements.txt
```
rchen152, line 51: nit: Please format this as:

\`\`\`shell
pip install -r requirements.txt
\`\`\`

to match the rest of our docs.

CONTRIBUTING.md
```diff
@@ -44,6 +44,13 @@ dependencies.
    Hence, to run these tests, you will need Python 2.7 and Python 3.x
    (preferably 3.6+) interpreters installed on your system.
 
+Required Python packages are listed in [requirements.txt](requirements.txt) file
```
rchen152, line 47: Suggested rewording for clarity:

Required Python packages are listed in **the** [requirements.txt](requirements.txt) file in this repository. **They can be installed** ~This can be done~ with pip with the following command:.

rchen152 (changes requested):
Thanks! A couple of minor comments.

CONTRIBUTING.md
josalmi: Done. Would you like me to also update the rest of the shell parts in this file while I'm at it? I was trying to follow the style in this particular file.

rchen152: Ah, thanks for pointing that out; I hadn't noticed. I'll update the rest of the formatting in a follow-up change.

rchen152 (approved):
Thanks again!

PR #696, opened 2020-10-03
Add Dockerfile for development

josalmi:
The Dockerfile includes all the dependencies needed to make pytype run in Docker and hopefully makes contributing easier for some.


rchen152 (approved):
I'm not too familiar with Dockerfiles, but this looks reasonable to me. Thank you for the contribution!

PR #707, opened 2020-10-12
Prepare to use pybind11.

Solumin:
- adds pybind11 as a submodule.
- disables cfg.py and builds cfg.so instead.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,6 +7,3 @@
 [submodule "cpython"]
 	path = cpython
 	url = https://github.com/python/cpython.git
-[submodule "pybind11"]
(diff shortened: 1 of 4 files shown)


.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "cpython"]
 	path = cpython
 	url = https://github.com/python/cpython.git
+[submodule "pybind11"]
```
rchen152, line 10: Why does pybind11 need to be added as a submodule?

Solumin: I thought it needed to be, but pip installing it via requirements.txt is actually sufficient.

Issue #706, opened 2020-10-12
merge-pyi does not support "import collections.abc as abc" style imports in stub files


ShaneHarvey:
merge-pyi does not support "import collections.abc as abc" style imports in stub files:

```
$ cat import-bug.py
import collections.abc as abc
$ cat import-bug.pyi
import collections.abc as abc
$ pytype --version
2020.10.08
$ merge-pyi import-bug.py import-bug.pyi
Traceback (most recent call last):
  File "/Users/shane/pymongo-pycharm-3.8/bin/merge-pyi", line 8, in <module>
    sys.exit(main())
  File "/Users/shane/pymongo-pycharm-3.8/lib/python3.8/site-packages/pytype/tools/merge_pyi/main.py", line 82, in main
    annotated_src = merge_pyi.annotate_string(args, py_src, pyi_src)
  File "/Users/shane/pymongo-pycharm-3.8/lib/python3.8/site-packages/pytype/tools/merge_pyi/merge_pyi.py", line 936, in annotate_string
    parsed_pyi = Pyi.parse(pyi_src)
  File "/Users/shane/pymongo-pycharm-3.8/lib/python3.8/site-packages/pytype/tools/merge_pyi/merge_pyi.py", line 763, in parse
    pkg, names = cls.parse_top_import(match_results)
  File "/Users/shane/pymongo-pycharm-3.8/lib/python3.8/site-packages/pytype/tools/merge_pyi/merge_pyi.py", line 834, in parse_top_import
    pkg_out = cls._parse_import_alias(list(pkg[0].leaves()))
  File "/Users/shane/pymongo-pycharm-3.8/lib/python3.8/site-packages/pytype/tools/merge_pyi/merge_pyi.py", line 801, in _parse_import_alias
    assert [leaf.type for leaf in leaves] == [token.NAME] * 3
AssertionError
```

rchen152:
Thanks for the report!

PR #713, opened 2020-10-16
merge-pyi support for "import collections.abc as abc" style imports

ShaneHarvey:
Fixes https://github.com/google/pytype/issues/706.


rchen152 (approved):
Thanks for this contribution! Sorry for the review delay - this all looks good to me.

PR #714, opened 2020-10-16
Finishing touches for using pybind11.

Solumin:


henryiii:
Suggestions: you should put pybind11 in your pyproject.toml, or use it as a submodule, probably not both. You can use the helpers inplace from the submodule, or just import then if it's in your pyproject.toml (since pip 10+ ensures you have everything in pyproject.toml temporarily installed when you are building a wheel). You don't need it in requirements.txt, as it's used at build time, not runtime (but it doesn't hurt, and makes development a tad easier).

henryiii:
Which method would you like? (since you already have submodules, I'd vote for that one ;) ) I can make a PR with a few suggestions if you let me know which one is preferred.

PR #716, opened 2020-10-16
fix: simplifiy pybind11 usage

henryiii:
This is a suggestion for using pybind11 via submodule. pyproject.toml is more elegant, but does require Pip 10+. I can switch to that if requested. Follow up on #714. I don't know all the details of building from scratch (It wants bison 3), but this should be close, I think. CC @rwgk .

I did not include ParallelCompile, but that might also be useful, I think.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools>=40.8.0", "ninja", "wheel"]
+requires = ["setuptools>=40.8.0", "ninja", "wheel", "pybind11==2.6.0rc3"]
 build-backend = "setuptools.build_meta"
--- a/setup.py
+++ b/setup.py
@@ -10,18 +10,18 @@ import re
 import shutil
 import sys
 
-from setuptools import setup  # pylint: disable=g-multiple-import
+from setuptools import setup
 
 # Path to directory containing setup.py
 here = os.path.abspath(os.path.dirname(__file__))
 
 # Get the pybind11 setup helpers
+#
+# This is appended, so if already available in site-packages, that is used
+# instead
 sys.path.append(os.path.join(here, "pybind11"))
-try:
-    from pybind11.setup_helpers import Pybind11Extension
-except ImportError:
-    print("pybind11 missing, probably a source checkout without submodules?")
-    raise
+
+from pybind11.setup_helpers import Pybind11Extension  # pylint: disable=g-import-not-at-top
 
 del sys.path[-1]
 
--- a/setup.py
+++ b/setup.py
@@ -10,10 +10,20 @@ import re
 import shutil
 import sys
 
-from setuptools import setup, Extension  # pylint: disable=g-multiple-import
+from setuptools import setup  # pylint: disable=g-multiple-import
 
 # Path to directory containing setup.py
 here = os.path.abspath(os.path.dirname(__file__))
+
+# Get the pybind11 setup helpers
+sys.path.append(os.path.join(here, "pybind11"))
+try:
+    from pybind11.setup_helpers import Pybind11Extension
+except ImportError:
+    print("pybind11 missing, probably a source checkout without submodules?")
+    raise
+
+del sys.path[-1]
 
 try:
   # The repository root is not on the pythonpath with PEP 517 builds
@@ -29,44 +39,27 @@ except ImportError:
 
 def get_parser_ext():
   """Get the parser extension."""
+  extra_compile_args = []
   # We need some platform-dependent compile args for the C extensions.
   if sys.platform == 'win32':
     # windows does not have unistd.h; lex/yacc needs this define.
-    extra_compile_args = ['-DYY_NO_UNISTD_H']
-    extra_link_args = []
-  elif sys.platform == 'darwin':
-    # clang on darwin requires some extra flags, which gcc does not support
-    extra_compile_args = ['-std=c++11', '-stdlib=libc++']
-    extra_link_args = ['-stdlib=libc++']
-  else:
-    extra_compile_args = ['-std=c++11']
-    extra_link_args = []
-  return Extension(
+    extra_compile_args.append('-DYY_NO_UNISTD_H')
+
+  return Pybind11Extension(
       'pytype.pyi.parser_ext',
       sources=[
           'pytype/pyi/parser_ext.cc',
           'pytype/pyi/lexer.lex.cc',
           'pytype/pyi/parser.tab.cc',
       ],
+      cxx_std=11,
       extra_compile_args=extra_compile_args,
-      extra_link_args=extra_link_args
   )
 
 
 def get_typegraph_ext():
   """Generates the typegraph extension."""
-  if sys.platform == 'win32':
-    # At this time, windows requires no additional flags.
-    extra_compile_args = []
-    extra_link_args = []
-  elif sys.platform == 'darwin':
-    # clang on darwin requires some extra flags, which gcc does not support
-    extra_compile_args = ['-std=c++11', '-stdlib=libc++']
-    extra_link_args = ['-stdlib=libc++']
-  else:
-    extra_compile_args = ['-std=c++11']
-    extra_link_args = []
-  return Extension(
+  return Pybind11Extension(
     'pytype.typegraph.cfg',
     sources=[
       "pytype/typegraph/cfg.cc",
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -15,6 +15,18 @@
 # Path to directory containing setup.py
 here = os.path.abspath(os.path.dirname(__file__))
 
+DIR = os.path.abspath(os.path.dirname(__file__))
```
henryiii, line 18: I can't read, apparently...

Solumin:
I didn't see this in time for #714, sorry! Can you clarify how pyproject.toml would work with using pybind11's CMake rules? We're using `pybind11_add_module` in pytype/typegraph/CMakeLists.txt.

Additionally, this should merge on the google_sync branch, since that's adding the file that depends on pybind11.

henryiii:
For using CMake, it's probably better to use the submodules (especially since you already are using them for other things) - if you ever allow a CMake run _without_ running it from setup.py, you can't force the items in pyproject.toml; users would have to install pybind11 manually (but if they already do with requirements.txt, for example, maybe that's not an issue?)

If you use FindPython, it's easy:

```cmake
find_package(Python COMPONENTS Interpreter Development)

execute_process(
    COMMAND "${Python_EXECUTABLE}" "-c"
            "import pybind11; print(pybind11.get_cmake_dir()')"
    OUTPUT_VARIABLE PYBIND11_ROOT
    OUTPUT_STRIP_TRAILING_WHITESPACE)

find_package(pybind11 CONFIG REQUIRED)
```

Feel free to check for errors on the `execute_process` command to make it more user friendly if something goes wrong.

If you use the classic FindPythonLibs, it's harder, since pybind11 modifies the Python search so heavily. If you found PythonInterp and PythonLibs already, or got them from the setup.py run, etc, then just run `PYTHON_EXECUTABLE` instead of `Python_EXECUTABLE`, but otherwise the same as above.

henryiii:
You could also do a hybrid, where it is used from the environment if available, or from the submodule if present, so you don't have to put it into the SDist. But you'd have to keep them in sync manually.

Solumin:
> You could also do a hybrid, where it is used from the environment if available, or from the submodule if present, so you don't have to put it into the SDist. But you'd have to keep them in sync manually.

This is more or less what I did in #715 in setup.py, where pybind11 (2.6.0rc2) is in requirements.txt, then setup.py tries to import it. But using the code from this PR should work better.

Unfortunately, I'm not very familiar with our build system, so I lack the knowledge to really comment on your previous comment. Besides agreeing that sticking with a submodule is fine, at least.

Solumin:
If the submodule overrides, the site-packages version will never be used and it doesn't need to be included in requirements.txt, right?

If site-packages overrides, I think that will give better stability for users, because they'll get pybind11 bug fixes faster. That's worth it to me.
On the other hand, we'll have more control over the submodule update schedule, which makes it easier for us to make sure pytype continues to build, I think. We can more easily keep the pybind version used on GitHub in sync with what's being used in Google.

I think the site-packages version should take precedence then.

henryiii:
Okay, I can try that.

> the site-packages version will never be used

Does this/ will this ever support building from SDist? If that's the case, then the submodule doesn't have to be grafted into your SDist, and therefore compiling from SDist would not have the submodule. Also someone could forget to clone the submodules. (not sure how this, especially SDist, would handle CPython, etc).

When you are building, the items in requirements.txt or the user's environment aren't used - pip makes a new environment with just the pyproject.toml requirements and builds there (unless you pass `--no-build-isolation`, which is rare, mostly only used when making conda packages). So for practical purposes, it will still be pinned by pyproject.toml.

By the way, that's a good reason to prioritize on site packages - conda-forge may ship a modified pybind11 and you could take advantage of that.

Solumin:
We do in fact use sdist when pushing a release to pypi: https://github.com/google/pytype/blob/master/build_scripts/release.py#L92
In that case, site packages is the way to go, if I'm following what you're saying.

setup.py
```diff
@@ -10,11 +10,21 @@
...
 
+# Get the pybind11 setup helpers
+#
+# This is appended, so if already available in site-packages, that is used
+# instead
+sys.path.append(os.path.join(here, "pybind11"))
+
+from pybind11.setup_helpers import Pybind11Extension  # pylint: disable=g-import-not-at-top
+
+del sys.path[-1]
```
Solumin, line 26: Can you add a comment explaining why this is necessary?

setup.py
```diff
@@ -10,11 +10,21 @@
...
 # Path to directory containing setup.py
 here = os.path.abspath(os.path.dirname(__file__))
 
+# Get the pybind11 setup helpers
+#
+# This is appended, so if already available in site-packages, that is used
+# instead
+sys.path.append(os.path.join(here, "pybind11"))
+
+from pybind11.setup_helpers import Pybind11Extension  # pylint: disable=g-import-not-at-top
```
Solumin, line 24: Can this import be placed before the setuptools import? Or does it need to be later?

henryiii:
That's odd, I'm triggering

```
setup.py:24:0: C0413: Import "from pybind11.setup_helpers import Pybind11Extension" should be placed at the top of the module (wrong-import-position)
```

But I have:

```python
from pybind11.setup_helpers import Pybind11Extension  # pylint: disable=g-import-not-at-top
```

setup.py
```diff
@@ -29,44 +39,27 @@
...
-    # clang on darwin requires some extra flags, which gcc does not support
-    extra_compile_args = ['-std=c++11', '-stdlib=libc++']
-    extra_link_args = ['-stdlib=libc++']
-  else:
-    extra_compile_args = ['-std=c++11']
-    extra_link_args = []
-  return Extension(
+    extra_compile_args.append('-DYY_NO_UNISTD_H')
+
+  return Pybind11Extension(
```
Solumin, line 48: The parser_ext doesn't use pybind11. I assume this will still work fine, but should be remain a normal Extension anyway?

Solumin:
> That's odd, I'm triggering
> 
> ```
> setup.py:24:0: C0413: Import "from pybind11.setup_helpers import Pybind11Extension" should be placed at the top of the module (wrong-import-position)
> ```
> 
> But I have:
> 
> ```python
> from pybind11.setup_helpers import Pybind11Extension  # pylint: disable=g-import-not-at-top
> ```

Moving the import up will help, or using `# pylint: disable=g-import-not-at-top,wrong-import-position` might fix it, I think.

setup.py
henryiii: This is just a lightweight subclass of Extension that adds pybind11's flags, the cxx_std option, and include dir. It was carefully designed to work anywhere a normal Extension works and does not require a special `build_ext`.

setup.py
henryiii: Sure, that's fine.

setup.py
henryiii: It's just tidy. You could leave `$PWD/pybind11` in the path if you prefer.

henryiii: I'll remove it, since you don't clean up the build_scripts addition below either.

Solumin (approved):

Solumin:
This looks great, thank you so much for doing this!

pyproject.toml
```diff
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools>=40.8.0", "ninja", "wheel"]
+requires = ["setuptools>=40.8.0", "ninja", "wheel", "pybind11==2.6.0rc3"]
```
martindemello, line 2: why `==` rather than `>=` for the requirement here?

henryiii: pyproject.toml's usually pin requirements fairly tightly; it can be used to create a completely reproducible build. Since the items here are _always_ downloaded and installed into the temporary build environment, there's no need to make the requirements too wide. Though it could be loosened up if you want. I think 2.6.0 will be out Monday-ish; then you can change this. I would probably not do more than `"pybind11~=2.6"`, though. (The tilde means the final digit, the 6 in this case, can be that or higher)

henryiii: My recommendation would be to keep it pinned to match the submodule, though.

I realize that I forgot to drop this from the MANIFEST, will do that.

henryiii: Here's an example of when it is very important to pin exactly: https://scikit-hep.org/developer/packaging#special-additions-numpy

martindemello: thanks, makes sense!

PR #726, opened 2020-11-04
Fix pip install errors.

rchen152:
I was getting errors about cfg_logging.h being missing when trying to
test a new release; this change seems to fix the issue.

I'm still seeing some odd dynamic import failure, but python3.7 on my
machine also appears to be half-broken right now, so I have no idea if
the issue is with the pytype code or my environment.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,4 +2,4 @@ include LICENSE
 include pyproject.toml
 include pytype/pyi/*
 include pytype/test_data/*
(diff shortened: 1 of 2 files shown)


rchen152:
I fixed my Python 3.7 installation, and the pytype install was still broken, so I've pushed a commit to also fix the dynamic import failure.

Solumin (approved):

PR #742, opened 2020-11-28
Updated the Tests badge to show build status from travis-ci.com

mraarif:
travis-ci.org is going to shut down on **December 31st, 2020**, repository should be migrated to travis-ci.com as mentioned here in docs: https://docs.travis-ci.com/user/migrate/open-source-repository-migration

Once we've migrated to travis-ci.com, we would need to update the tests status badge to show status from travis-ci.com.

PR #743, opened 2020-11-30
Added GitHub workflow for CI (use GitHub Actions) to replace travis

mraarif:
As noted in the previous PR(#742 ), this repository still seems to be using travis-ci.org, and travis-ci.com's performance seems to be degraded in the past few weeks, since all the other checks are using GitHub actions, this PR adds a new GitHub workflow to run CI checks to replace Travis.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
...
+
+    runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.experimental }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04]
+        python-version: ['3.6','3.7','3.8']
+        experimental: [false]
+        include:
+          - python-version: '3.9-dev'
```
rchen152, line 20: nit: Could you change the order of this section to
```
- os
  python-version
  experimental
```
to match lines 16-18? Otherwise, the 3.9-dev build has a different name from the others (https://github.com/google/pytype/pull/743/checks), which is slightly confusing.

rchen152 (changes requested):
This looks great except for one major sticking point - the 3.9-dev check failing causes the entire workflow to be marked as failed. GitHub Actions seems to be missing an equivalent to Travis CI's "allow_failures" option (https://github.com/actions/toolkit/issues/399). This seems like it would be quite inconvenient (can't tell whether a PR passed tests or not, master branch appears to always be broken). I looked through the GitHub Actions issue but didn't see any obvious way to simulate allow_failures - do you have any ideas? I'd be okay with the 3.9 check always reporting success, regardless of whether it actually succeeded, as long as the logs are available somewhere.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,51 @@
...
+    - name: Install Dependencies
+      run: pip install -r requirements.txt
+
+    - name: Run Tests
+      if: matrix.python-version != '3.9-dev'
+      run: python build_scripts/travis_script.py
+
+    - name: Run Tests for python 3.9
+      if: matrix.python-version == '3.9-dev'
+      run: python build_scripts/travis_script.py || exit 0
```
mraarif, line 51: this will make sure the build remains green if it fails while running on python 3.9 only and run normally otherwise for other python versions

README.md
```diff
@@ -1,4 +1,4 @@
-[![Tests](https://travis-ci.org/google/pytype.svg?branch=master)](https://travis-ci.org/google/pytype)
+![CI](https://github.com/google/pytype/workflows/CI/badge.svg?branch=master)(https://github.com/google/pytype/actions)
```
rchen152, line 1: Sorry I didn't catch this the first time around, but I believe you need an extra pair of brackets, like so:

<pre><b>[</b>CI</pre>

Right now, I'm seeing the second url displayed to the right of the badge: https://github.com/google/pytype/blob/98417a3dd47ef5947d7614cec87b8ccfa342140e/README.md

mraarif: my bad, didn't have a good markdown editor I guess.

rchen152 (approved):
Looks great, thanks! I'll merge this once the tests pass.

Issue #759, opened 2020-12-13
RFE: add a .gitignore to output dir


scop:
Would be good to automatically add a `.gitignore` to the output dir when creating it, e.g. with contents
```gitignore
# Automatically created by pytype
*
```
This way people wouldn't need to manually add it to their upper level gitignores. Many tools already do this for their similar dirs, mypy and pytest to name a couple.

rchen152:
That sounds totally reasonable to me! I must confess, I didn't know until now that you could add a `.gitignore` to a subdirectory.

UrielMaD:
Hi @rchen152 I'd like to take this issue :)

rchen152:
Sweet, thank you for your interest in contributing to pytype! The path to the output directory is available as the "output" attribute on the configuration object (`conf`); it would probably make the most sense to add the `.gitignore` right after the directory is created here: https://github.com/google/pytype/blob/51e930ca7d29a73ede531129111fb600e761c8cc/pytype/tools/analyze_project/main.py#L76.

UrielMaD:
Hello @rchen152 I just send a PR for adding the git ignore file to the output dir, I'm wondering if there's an specific configuration for the file that will be created.

The current configuration that I set is just what @scop suggested as example

rchen152:
`*` seems fine to me; I don't think there's anything we specifically want to not be excluded.

UrielMaD:
Nice. Let me know if there's something else that needs to be implemented!

rchen152:
@UrielMaD Thank you for implementing this feature! It'll be in the next pytype release (first or second week of January).

PR #770, opened 2020-12-23
Automatically added git ignore file to sub directory

UrielMaD:
This PR automatically creates a .gitignore file to the output directory of pytype/tools/analyze_project/main.py

Issue #759

### DIFF
--- a/pytype/tools/analyze_project/main.py
+++ b/pytype/tools/analyze_project/main.py
@@ -74,7 +74,7 @@ def main():
   logging.info('Source tree:\n%s',
                importlab.output.formatted_deps_list(import_graph))
   tool_utils.makedirs_or_die(conf.output, 'Could not create output directory')
-  f = open(conf.output + ".pytype/.gitignore", "w")
+  f = open(conf.output, ".gitignore", "w")
   f.write("# Automatically created by pytype\n*")
   deps = pytype_runner.deps_from_import_graph(import_graph)
   runner = pytype_runner.PytypeRunner(conf, deps)
--- a/pytype/tools/analyze_project/main.py
+++ b/pytype/tools/analyze_project/main.py
@@ -74,6 +74,9 @@ def main():
   logging.info('Source tree:\n%s',
                importlab.output.formatted_deps_list(import_graph))
   tool_utils.makedirs_or_die(conf.output, 'Could not create output directory')
+  f = open(conf.output + ".gitignore", "w")
+  f.write("# Automatically created by pytype\n*")
+  f.close()
   deps = pytype_runner.deps_from_import_graph(import_graph)
   runner = pytype_runner.PytypeRunner(conf, deps)
   return runner.run()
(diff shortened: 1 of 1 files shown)


UrielMaD:
@googlebot I signed it!

pytype/tools/analyze_project/main.py
```diff
@@ -74,6 +74,9 @@ def main():
   logging.info('Source tree:\n%s',
                importlab.output.formatted_deps_list(import_graph))
   tool_utils.makedirs_or_die(conf.output, 'Could not create output directory')
+  f = open(conf.output + ".gitignore", "w")
```
rchen152, line 77: This will create a file called ".pytype.gitignore" (assuming the output dir is .pytype/). If you use `os.path.join(conf.output, '.gitignore')`, that should create ".pytype/.gitignore", which is what you want.

rchen152, line 77: I'd suggest using a context manager, so you don't have to remember to call `.close()`:
```
with open(...) as f:
  f.write(...)
  # no need to call f.close() here, since the context manager will clean up after you!
```

rchen152 (changes requested):

pytype/tools/analyze_project/main.py
UrielMaD: Thanks for your review! I will modify it right away!

UrielMaD: Code is updated :D

rchen152: Thanks!

pytype/tools/analyze_project/main.py
```diff
@@ -74,6 +74,8 @@ def main():
   logging.info('Source tree:\n%s',
                importlab.output.formatted_deps_list(import_graph))
   tool_utils.makedirs_or_die(conf.output, 'Could not create output directory')
+  with open(conf.output + '/.gitignore', 'w') as f:
```
rchen152, line 77: Please use `os.path.join(conf.output, 'gitignore')` instead of `conf.output + '/.gitignore'`. (You'll also have to import os at the top of the file.) Using os.path.join avoids hard-coding a platform-specific separator and generates the correct path regardless of whether conf.output ends with a /.

rchen152 (changes requested):
This looks pretty good now, just one more minor comment.

pytype/tools/analyze_project/main.py
UrielMaD: Thanks for your time @rchen152, changes are updated!

rchen152 (approved):

pytype/tools/analyze_project/main.py
rchen152: Thank you! I'll import this change once the tests pass. (You'll see the PR being closed, rather than merged, but you'll be credited as the author in the commit history.)

PR #811, opened 2021-01-19
Get the line number of first line in function body

### LINKED ISSUE #783: Incorrect raising of Function type comments cannot be used with annotations [redundant-function-type-comment]
Python version: 3.8.6
Pytype version: 2020.12.23
Tested in python:3.8.6-buster, OSX virtual env

I have provided a minimal script which helps explain the issue. I am getting an error where by commenting or uncommenting a line of code is causing `redundant-function-type-comment` errors to be thrown. Which appears to be a little inconsistent (and incorrect).

❌ The following code, run with the command will result in the error:
```
root@a93364f24f3b:/opt/app# pytype test.py
Computing dependencies
Analyzing 1 sources with 0 local dependencies
ninja: Entering directory `/opt/app/.pytype'
[1/1] check test
FAILED: /opt/app/.pytype/pyi/test.pyi 
/venv/bin/python -m pytype.single --disable pyi-error --imports_info /opt/app/.pytype/imports/test.imports --module-name test -V 3.8 -o /opt/app/.pytype/pyi/test.pyi --analyze-annotated --nofail --quick /opt/app/test.py
File "/opt/app/test.py", line 7: Function type comments cannot be used with annotations [redundant-function-type-comment]

```
Code - test.py
```
from typing import Dict


def a() -> Dict[str, str]:
    # some_variable = ''
    # Array is going to be encoded as a string, we do this because of legacy
    cast_type: Dict[str, str] = {
        'boolean': 'string',
        'Float': 'float',
        'Array': 'string',
    }

    return cast_type


print(a())
```

✅ Now if i were to modify the code in the following way (uncomment the `some_variable`). Pytpye passes the check as valid.

Command:
```
root@a93364f24f3b:/opt/app# pytype test.py
Computing dependencies
Analyzing 1 sources with 0 local dependencies
ninja: Entering directory `/opt/app/.pytype'
[1/1] check test
Leaving directory '/opt/app/.pytype'
Success: no errors found
```

Code - test.py
```
from typing import Dict


def a() -> Dict[str, str]:
    some_variable = ''
    # Array is going to be encoded as a string, we do this because of legacy
    cast_type: Dict[str, str] = {
        'boolean': 'string',
        'Float': 'float',
        'Array': 'string',
    }

    return cast_type


print(a())
```

Making the following single changes to the original code:
✅  Removing the type attached to `cast_type: Dict[str, str] = {` results in a pass.
✅ Adding a type to `some_variable -> some_variable: str` results in a passes.
❌ Reming `some_variable` and removing code comment results in a fail.

parampatidar:
Resolves #783

### DIFF
--- a/pytype/blocks.py
+++ b/pytype/blocks.py
@@ -340,7 +340,8 @@ class CollectAnnotationTargetsVisitor:
         fn_code = code.co_consts[code_op.arg]
         if not _is_function_def(fn_code):
           continue
-        end_line = fn_code.co_code[0].line  # First line of code in body.
+        # First line of code in body.
+        end_line = min(op.line for op in fn_code.co_code)
         self.make_function_ops[op.line] = (end_line, op)
       elif (isinstance(op, STORE_OPCODES) and
             op.line not in self.make_function_ops):
--- a/pytype/tests/py3/test_type_comments.py
+++ b/pytype/tests/py3/test_type_comments.py
@@ -39,5 +39,18 @@ class Py3TypeCommentTest(test_base.TargetPython3FeatureTest):
         v: int = None  # type: str  # ignored-type-comment
     """)
 
+  def test_first_line_of_code(self):
+    self.Check("""
+      from typing import Dict
+      def f() -> Dict[str, int]:
+        # some_var = ''
+        # something more
+        cast_type: Dict[str, int] = {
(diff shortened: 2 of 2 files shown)


rchen152 (approved):
Thank you for the contribution!

Issue #791, opened 2021-01-08
Support PEP 593: Flexible Function and Variable Annotations


rchen152:
PEP: https://www.python.org/dev/peps/pep-0593/
This looks like it entails adding an Annotated type that returns its first parameter and throws away the second.

This would be a good first issue for someone interested in learning how to support new typing features in pytype, since it requires going through all the steps described [here](https://google.github.io/pytype/developers/features.html) for adding a new feature, but none of the steps in this case should be particularly tricky.

mlevesquedion:
I would be interested in implementing this feature.

rchen152:
Great! I think implementing this will require:
* Adding a definition for `Annotated` to the typing stubs in `pytype/stubs/builtins/`
* Adding an implementation of `Annotated` to `pytype/overlays/typing_overlay.py`. You'll probably want to inherit from `abstract.AnnotationClass` and override `_build_value` (similar to what the implementation of `Union` does: https://github.com/google/pytype/blob/644f04dc7cdd3ccb836e5758f175b12028fbb888/pytype/overlays/typing_overlay.py#L46).
* Adding an alias for `Annotated` in `pytype/overlays/typing_extensions_overlay.py`, so it can be used in versions prior to 3.9.

Full instructions are here: https://google.github.io/pytype/developers/features.html. (Not all of the listed steps apply in this case.) This documentation is new, so please ask if anything doesn't make sense!

mlevesquedion:
Sounds good. Thanks a lot for the directions!

rchen152:
@mlevesquedion Thanks again for implementing this! It'll be in the next pytype release (tomorrow or Friday, if nothing unexpected comes up).

mlevesquedion:
Thank you for your help! :slightly_smiling_face: I look forward to contributing again if I get the chance.

PR #819, opened 2021-01-26
Implement basic support for PEP-593

mlevesquedion:
This PR implements basic support for PEP-593 (relevant issue: #791).

This is my first time contributing, so I have a few questions:
* I saw instructions about debugging in the [developer docs](https://google.github.io/pytype/developers/process.html), but I couldn't get them to work. I tried adding `import pdb; pdb.set_trace()` in my tests, but when I run the tests using `python3.8 build_scripts/run_tests.py pytype.tests.test_annotated -v`, `pdb` quits immediately with a `BdbQuit` exception. AFAIU, this is because the test is running in `ninja`, i.e. non-interactively, so `pdb` can't do its thing. What is the recommended workflow for debugging tests?
* Do I need to add stubs in `stubs/builtins/{2,2and3}` even though this is strictly a Python 3 feature? (Actually, do I need to add stubs at all, since there is an overlay?)

I have one additional question concerning the PEP itself:
According to the PEP, the type `typing.Annotated[int, "foo"]` is not the same as the type `typing.Annotated[int, "bar"]`. My question is: does `pytype` care about type equality? Currently, the code merely drops all annotations, so `typing.Annotated[int, ...]` becomes just `int`, e.g.


pytype/stubs/builtins/3/builtins.pytd
```diff
@@ -3,7 +3,7 @@
 
 
 from typing import (
-    AbstractSet, Any, AnyStr, AsyncGenerator, Awaitable, BinaryIO, Callable,
+    AbstractSet, Annotated, Any, AnyStr, AsyncGenerator, Awaitable, BinaryIO, Callable,
```
rchen152, line 6: Annotated doesn't need to be imported here. (The imports here are for typing members that are needed to type-annotate builtins.)

pytype/tests/py3/CMakeLists.txt
```diff
@@ -9,6 +9,16 @@ py_test(
     pytype.tests.test_base
 )
 
+py_test(
+  NAME
+    test_annotated
```
rchen152, line 14: nit: Could you call this test_typing_annotated? pytype has a lot of functional tests, and it's hard to tell at first glance what "test_annotated" might be testing.

pytype/tests/py3/test_annotated.py
```diff
@@ -0,0 +1,65 @@
+"""Tests for PEP-593 typing.Annotated types."""
+
+from pytype.tests import test_base
+
+
+class AnnotatedTest(test_base.TargetPython3BasicTest):
```
rchen152, line 6: This test class should inherit from TargetPython3*Feature*Test. For the GitHub tests it doesn't matter, but internally, TargetPython3BasicTest tests are run in both Python 3 and Python 2 (using a patched interpreter that has type annotations backported).

pytype/tests/py3/CMakeLists.txt
```diff
@@ -9,6 +9,16 @@ py_test(
...
+  SRCS
+    test_annotated.py
+  DEPS
+    pytype.utils
```
rchen152, line 18: The pytype.utils dependency appears to be unneeded.

rchen152 (changes requested):
This is fantastic, thank you! In particular, I appreciate you including such an extensive test suite.

* For debugging tests, I usually copy the test code into a file, rebuild the `pytype-single` executable from the source code after inserting breakpoint(s), and then run `pytype-single` on the file. You can install `pytype-single` from source either by doing `pip install -e .` in the repo's top-level directory or by running `python build_scripts/build.py` and then using `out/bin/pytype-single`.
* I agree, only changing the Python 3 stubs should be fine. We do still need a basic definition for Annotated in typing.pytd; otherwise pytype would complain about Annotated not existing if it tries to parse a .pyi file containing it. Just `class Annotated: ...` is probably fine for now, since there isn't yet a way to express that Annotated takes a variable number of type parameters.

rchen152:
> According to the PEP, the type typing.Annotated[int, "foo"] is not the same as the type typing.Annotated[int, "bar"]. My question is: does pytype care about type equality? Currently, the code merely drops all annotations, so typing.Annotated[int, ...] becomes just int, e.g.

It should be fine to drop the annotations. pytype doesn't really try to handle typing metaprogramming anyway.

pytype/tests/py3/CMakeLists.txt
mlevesquedion: Done.

pytype/tests/py3/test_annotated.py
mlevesquedion: Done, thanks for explaining the difference!

pytype/tests/py3/CMakeLists.txt
mlevesquedion: Of course. Done.

pytype/stubs/builtins/3/builtins.pytd
mlevesquedion: Thanks for explaining. I've removed the import and added a stub instead.

mlevesquedion:
Thank you very much for reviewing and providing many helpful explanations!

Thanks to your pointers I was able to run `pdb` in `pytype-single`. I couldn't run it earlier due to import errors when running `out/bin/pytype-single`. I thought `pytype-single` was some kind of binary so I wasn't sure how to debug the errors, but turns out it's a script with a shebang that points to ` /usr/bin/env python`, which on my machine is Python 2 :upside_down_face:.

pytype/stubs/builtins/3/builtins.pytd
```diff
@@ -13,6 +13,9 @@ from typing import (
 if sys.version_info >= (3, 8):
     from typing import SupportsIndex
 
+# typing.Annotated, added in Python 3.9
+class Annotated: ...
```
rchen152, line 17: This should be in pytype/stubs/builtins/3/*typing*.pytd (not builtins.pytd). I also suggest putting this inside a `sys.version_info >= (3, 9):` block, since the class isn't available prior to 3.9.

rchen152 (changes requested):
Thanks! Just one more comment.

pytype/stubs/builtins/3/builtins.pytd
mlevesquedion: Gotcha. Done.

rchen152 (approved):

Issue #493, opened 2020-01-18
False positive `str.isascii` missing attribute


AdamGleave:
Python 3.7 added [str.isascii](https://docs.python.org/3.7/library/stdtypes.html#str.isascii) to the standard library, but `pytype` doesn't seem to know about it yet.

For example, checking the following code:

```
def main():                                                                                                                                                                  
    return 'foo'.isascii()


if __name__ == '__main__':
    print(main())
```

gives the following error:

```
$ pytype test.py                                                                                                                
Computing dependencies                                                                                                                                                       
Analyzing 1 sources with 0 local dependencies                                                                                                                                
ninja: Entering directory `/home/adam/dev/evaluating-rewards/.pytype'                                                                                                        
[1/1] check test
FAILED: /home/adam/dev/evaluating-rewards/.pytype/pyi/test.pyi 
/home/adam/dev/evaluating-rewards/venv/bin/python3.7 -m pytype.single --imports_info /home/adam/dev/evaluating-rewards/.pytype/imports/test.imports --module-name test -V 3.7 -o /home/adam/dev/evaluating-rewards/.pytype/pyi/test.pyi --analyze-annotated --nofail --quick /home/adam/dev/evaluating-rewards/test.py
File "/home/adam/dev/evaluating-rewards/test.py", line 2, in main: No attribute 'isascii' on str [attribute-error]

For more details, see https://google.github.io/pytype/errors.html#attribute-error.
ninja: build stopped: subcommand failed.
```

This is with:
```
$ pytype --version
2020.01.08
```

rchen152:
Looks like we need to add the attribute here: https://github.com/google/pytype/blob/ce086c5c6ef55364dac96a3c1e97a4a21b1aefeb/pytype/pytd/builtins/3/__builtin__.pytd#L208.

rchen152:
@Satyabhama-Reddy Thanks for fixing this issue! The fix will be in the next pytype release (targeted for later this week).

PR #825, opened 2021-01-29
Adding isascii attribute to str,bytes and bytearray

Satyabhama-Reddy:
Pull Request for Issue "False positive `str.isascii` missing attribute #493"

The documentation of python 3.7 indicates isascii() was added to bytes and bytearray as well. Hence adding those too.
https://docs.python.org/3/whatsnew/3.7.html
"""str, bytes, and bytearray gained support for the new isascii() method, which can be used to test if a string or bytes contain only the ASCII characters. (Contributed by INADA Naoki in bpo-32677.)"""


rchen152 (changes requested):
Thanks! pytype still supports Python 3.5 and 3.6, so could you add `if sys.version_info >= (3, 7):` guards before the definitions of isascii?

rchen152 (approved):

Issue #834, opened 2021-02-18
Installing with pipenv raises: "pytype/pyi/lexer.h:5:10: fatal error: Python.h: No such file or directory" on Linux Mint 20.1


Marto32:
Hello,

I'm trying to use pipenv to install pytype in a new environment and am running into the following error when running `pipenv install pytype` in that environment (python 3.9):

```
x86_64-linux-gnu-gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O2 -Wall -g -fstack-protector-strong -Wformat -Werror=format-security -g -fwrapv -O2 -g -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -I/home/michael/.local/share/virtualenvs/globo-68Lb6zOm/include -I/usr/include/python3.9 -c pytype/pyi/lexer.lex.cc -o build/temp.linux-x86_64-3.9/pytype/pyi/lexer.lex.o -std=c++11
    In file included from /usr/local/google/home/rechen/pytype/pytype/pyi/lexer.lex:17:
    pytype/pyi/lexer.h:5:10: fatal error: Python.h: No such file or directory
        5 | #include <Python.h>
          |          ^~~~~~~~~~
    compilation terminated.
    error: command '/usr/bin/x86_64-linux-gnu-gcc' failed with exit code 1
    ----------------------------------------
ERROR: Command errored out with exit status 1: /home/michael/.local/share/virtualenvs/globo-68Lb6zOm/bin/python -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-ufvelddb/pytype_59996eb61816442995882e12c4db05a3/setup.py'"'"'; __file__='"'"'/tmp/pip-install-ufvelddb/pytype_59996eb61816442995882e12c4db05a3/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-mypns2tg/install-record.txt --single-version-externally-managed --compile --install-headers /home/michael/.local/share/virtualenvs/globo-68Lb6zOm/include/site/python3.9/pytype Check the logs for full command output.
```

[...]

Issue #749, opened 2020-12-04
Python 3.9 is not supported


dundee:
Python 3.9 is out for some time now, it would be great if it's supported.

rchen152:
I recently did some work to start supporting 3.9, but we still need to implement new opcodes (https://github.com/google/pytype/blob/0790b3f0c1600d21a7d7a917eb408e29d772c365/pytype/vm.py#L3413) and fix test failures (https://github.com/google/pytype/runs/1488890229?check_suite_focus=true - the overall status is a lie, expand the "Run tests for python 3.9" section to see failures).

Added a "help wanted" label because I'd love help on this issue! Internally, we don't need 3.9 yet, so I've been chipping away at this in my spare time.

ShayNehmad-RecoLabs:
I'd like to add that we'd want this as well. How can I help if help is wanted? @rchen152 I'm not super familiar with pytype but with some direction, I could probably help.

kamahen:
Is there documentation on how to add a bytecode? I suppose I could figure it out by looking through the source (and maybe looking at https://github.com/nedbat/byterun), but if there's more information than contained in bytecode.md, that would be nice.
(No promises that I'll actually work on this, of course).

BTW, it would be nice to have type annotations in vm.py (and elsewhere) ... I presume pytype could be run on itself to generate thiese?

rchen152:
@ShayNehmad-RecoLabs The main thing standing in the way of 3.9 support is updating pytype to reflect changes in the Python bytecode. (pytype works by compiling source files to bytecode and then analyzing the bytecode.) pytype/vm.py contains a `byte_OPCODE_NAME` method for every bytecode operation, and the [documentation for the dis library](https://docs.python.org/3/library/dis.html) describes what each opcode does. What I typically do is add implementations for the new opcodes (I've put in empty stubs for the 3.9 ones at the bottom of vm.py: https://github.com/google/pytype/blob/0790b3f0c1600d21a7d7a917eb408e29d772c365/pytype/vm.py#L3413) based on the descriptions in the `dis` docs, then use test failures to figure out what behavior changes need to be made to existing opcode implementations.

@kamahen Have you seen the rest of the [developer guide](https://google.github.io/pytype/developers/index.html)? In particular, you might want to check out the other docs in the "Program Analysis" section.

Marto32:
Any updates on the timeline here? Would love to see this in 3.9.

rchen152:
[...]

PR #840, opened 2021-02-22
Allow pytype to be installed in Python 3.9.

rchen152:
For https://github.com/google/pytype/issues/749.

I manually ran the pytype tests with host version = 3.9 and target = 3.8
to confirm that, while pytype cannot yet analyze 3.9 code, it has no
trouble running under 3.9.

This change
(1) Ensures that users attempting to install pytype in 3.9 get the
latest version. Currently, they get version 2020.2.6, which is old and a
little broken (see, e.g., https://github.com/google/pytype/issues/834).
(2) Allows users to run pytype in 3.9 and analyze 3.5-3.8 code by
specifying --python_version. This is arguably not very useful, since
pytype won't be able to resolve pip-installed dependencies, but it may
be good enough for small, self-contained projects.

Attempting to run pytype in 3.9 without specifying a lower target version will still produce a 'Python versions > 3.8 are not yet supported' error.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,13 +20,14 @@ classifiers =
     Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
     Programming Language :: Python :: Implementation :: CPython
     Topic :: Software Development
 
 
 [options]
 zip_safe = False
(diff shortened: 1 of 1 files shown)


rchen152:
Fixed the CI errors and rebased, so hopefully the tests are happy now.

martindemello (approved):

PR #862, opened 2021-03-17
Fix dead pyi-examples link in user_guide.md

xiaoxiae:
Not sure what the meaning of /2 is, but clicking the link yields GitHub's 404. Removing it points to valid examples to `.pyi` files.


xiaoxiae:
@googlebot I signed it!

rchen152 (approved):
Thanks! typeshed used to have a stdlib/2/ directory for Python 2 stubs.

PR #903, opened 2021-05-03
Support max versions in typeshed VERSIONS file

srittau:
As discussed in python/typeshed#5304. Please note that this PR is mostly untested, but I hope CI passes.

### DIFF
--- a/pytype/pytd/typeshed.py
+++ b/pytype/pytd/typeshed.py
@@ -120,8 +120,8 @@ class Typeshed:
       match = re.fullmatch(r"(.+): (\d)\.(\d+)(?:-(?:(\d)\.(\d))?)?", line)
       assert match
       module, min_major, min_minor, max_major, max_minor = match.groups()
-      minimum = (int(major), int(minor))
-      maximum = (int(max_major), int(max_minor))
+      minimum = (int(min_major), int(min_minor))
+      maximum = ((int(max_major), int(max_minor))
                  if max_major is not None and max_minor is not None else None)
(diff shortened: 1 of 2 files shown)


rchen152:
Thanks! Looks pretty good to me, but let's see what the tests say. (Seems that workflows now need approval from a maintainer to run if a PR is from a first-time contributor...)

srittau:
Ah, found the tests. 😃 Will fix and provide tests for the new functionality.

pytype/pytd/typeshed.py
```diff
@@ -104,20 +104,26 @@ def _load_stdlib_versions(self):
...
 
     Returns:
-      A mapping from module name to (major, minor) Python version.
+      A mapping from module name to ((min_major, min_minor),
+      (max_major, max_minor)) Python version. The max tuple can be `None`.
     """
     _, text = self._load_file(os.path.join("stdlib", "VERSIONS"))
     versions = {}
     for line in text.splitlines():
-      match = re.fullmatch(r"(.+): (\d)\.(\d+)", line)
```
JelleZijlstra, line 117: While you're at it could you add support for skipping blank and comment lines here, like mypy does?

srittau:
Since the current tests also check against the real `VERSIONS` file, I think it's easier to wait for that file to be updated using the new syntax before adding tests.

JelleZijlstra (approved):

rchen152 (approved):
Thanks again! Importing this into our internal codebase now.

PR #906, opened 2021-05-04
fix handling of stdlib/@python2

JelleZijlstra:
Related to python/typeshed#5329: we will now set 2.7 as the minimum version in VERSIONS for libraries that existed in 2.7. I tested this by running pytype_test.py in typeshed.

Happily, fixing this also helped me find out that I got a few versions wrong in the VERSIONS file.


JelleZijlstra:
@googlebot I signed it!

rchen152 (approved):
Thanks!

Issue #785, opened 2021-01-07
Support PEP 604: Allow writing union types as X | Y


rchen152:
PEP: https://www.python.org/dev/peps/pep-0604/

Introduces new syntax that we should support as soon as possible so typeshed can start using it (https://github.com/python/typeshed/issues/4819).

commentator8:
Hi - i see the code seems to be checking if the context is 3.10 `if ctx.python_version >= (3, 10) and name == "__or__":` - we are using `from future import __annotations__` in python 3.9. Am i correct that it still wont work for 3.9 then?

martindemello:
Hi @commentator8 , that's right, the feature made it into python in 3.10 so we don't support it in earlier versions, even with `from future import __annotations__` or explicit string annotations.

commentator8:
Good to know. Thanks. I've recently disabled pytype for our company's pipelines and was hoping to bring it back.

PR #924, opened 2021-05-17
Add support for PEP 604 (Union |) in stub files

JelleZijlstra:
Part of #785, though not a complete fix because this doesn't allow the `|` syntax at runtime (I think).

I was pleasantly surprised by how easy this change was. I hope that doesn't mean I missed something.


rchen152:
Wow! That really is a shockingly easy change.

@martindemello Could you review this PR, since you wrote pyi/parser.py and are the most familiar with it?

martindemello:
looks great, thanks for the fix!

rchen152:
Great, I'll go ahead and import this PR. Thanks, Martin!

PR #944, opened 2021-06-14
Work with new typeshed structure

srittau:
It's now possible that third-party packages have a `@python2` directory and no stubs at top-level. This means they are Python 2-only, even if the corresponding `METADATA.toml` keys are not set.

There are likely more changes in the future.

See python/typeshed#5630 for details.

### DIFF
--- a/pytype/pytd/typeshed.py
+++ b/pytype/pytd/typeshed.py
@@ -144,9 +144,9 @@ class Typeshed:
     stubs/ contains type information for third-party packages. Each top-level
     directory corresponds to one PyPI package and contains one or more modules,
     plus a metadata file (METADATA.toml). If there are separate Python 2 stubs,
-    they live in an @python2 subdirectory. METADATA.toml
-    takes @python2 into account, so if a package has both foo.pyi and
-    @python2/foo.pyi, METADATA.toml will contain `python2 = True`.
+    they live in an @python2 subdirectory. If stubs support both Python 2 and
+    Python 3 without a separate @python2 directory, METADATA.toml will contain
+    `python2 = True`.
 
     Returns:
       A mapping from module name to a set of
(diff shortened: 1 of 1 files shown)


pytype/pytd/typeshed.py
```diff
@@ -144,32 +144,41 @@ def _load_third_party_packages(self):
...
       parts = third_party_file.split(os.path.sep)
       if parts[-1] == "METADATA.toml":  # {package}/METADATA.toml
         _, metadata_file = self._load_file(
             os.path.join("stubs", third_party_file))
         metadata[parts[0]] = toml.loads(metadata_file)
-      elif "@python2" not in parts:  # {package}/{module}
+      elif "@python2" in parts:  # {package}/@python2/{module}
+        py2 = True
+      elif "@python2" not in parts: # {package}/{module}
+        if parts[-1].endswith(".pyi"):
```
rchen152, line 168: Why is this check needed?

pytype/pytd/typeshed.py
```diff
@@ -144,32 +144,41 @@ def _load_third_party_packages(self):
...
+          py3 = True
         name, _ = os.path.splitext(parts[1])
         modules[parts[0]].add(name)
     packages = collections.defaultdict(set)
     for package, names in modules.items():
       for name in names:
-        # When not specified, packages are Python 3-only
-        if metadata[package].get("python2", False):
+        # When not specified and no @python2 directory exists, packages are
```
rchen152, line 175: I think this comment and the one in lines 179-180 are switched.

rchen152 (changes requested):
Thanks!

pytype/pytd/typeshed.py
srittau: There can be other files in the top-level directory, for example the `@tests` directory or a `README` file. The existence of these files shouldn't mean that the stubs support Python 3.

pytype/pytd/typeshed.py
srittau: I have reworked the comments. The logic should become a bit easier when we move all py2-only packages to `@python2` in typeshed.

pytype/pytd/typeshed.py
```diff
@@ -144,32 +144,41 @@ def _load_third_party_packages(self):
...
+    Python 3 without a separate @python2 directory, METADATA.toml will contain
+    `python2 = True`.
 
     Returns:
       A mapping from module name to a set of
       (package name, major_python_version) tuples.
     """
     metadata = {}
     modules = collections.defaultdict(set)
+    py2 = False  # does a @python2 directory exist
```
rchen152, line 157: Ah, not sure how I didn't notice this before, but I think py2 and py3 need to be defined inside the loop and stored in the 'modules' dict. (The reason this method is structured in this way - one loop over all the files in "stubs" and another to populate 'packages' - is that due to how we run pytype internally, file operations are expensive, so we want to do as few of them as possible.)

Sorry for the extra round trip.

pytype/pytd/typeshed.py
rchen152: Makes sense, thanks!

rchen152 (changes requested):

pytype/pytd/typeshed.py
srittau: Whoops. Not sure how I missed this. Will fix.

rchen152 (approved):
Thanks again!

PR #1053, opened 2021-11-17
Cancel the previous Build Wheels workflow when a new one starts.

rchen152:


rchen152:
I can't make heads or tails of the Build wheels failure here, and this wasn't an important change anywa, so I'm just going to drop it.

PR #1055, opened 2021-11-17
Only require `typed-ast` for Python < 3.8

lgeiger:
Since 2a6a1cebb13a71d5a2dda08ef2e260a81659b392 `pytype` will use the builtin `ast` module instead of `typed-ast` for Python 3.8 or newer. This PR removes the dependency from the requirements files for these newer Python versions.

### DIFF
--- a/pytype/tools/annotate_ast/main.py
+++ b/pytype/tools/annotate_ast/main.py
@@ -6,7 +6,13 @@ import sys
 from pytype.ast import debug
 from pytype.tools import arg_parser
 from pytype.tools.annotate_ast import annotate_ast
-from typed_ast import ast3
+
+# pylint: disable=g-import-not-at-top
+if sys.version_info >= (3, 8):
+  import ast as ast3
+else:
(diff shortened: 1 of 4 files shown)


rchen152 (changes requested):
Thanks for making this change! It looks like the tests are failing due to a new lint error: https://github.com/google/pytype/runs/4239855820?check_suite_focus=true. I believe that's a false positive caused by the linter not realizing that the stdlib `ast` library isn't being used pre-3.8; could you just add a `# pylint: disable=nexpected-keyword-arg` to the end of the problematic line?

lgeiger:
@rchen152 Thanks for checking CI. I ignored the linter warning in 3aecfef4a389551ca4deb30872fcc0eafff1b6d2

rchen152 (approved):
Looks good, thanks again! I'll merge this after doing today's release (so it has time to be imported and go through nightly testing), so it should be in the next release in ~a week.

PR #1091, opened 2021-12-19
Fix typos

kianmeng:


pytype/directors_test.py
```diff
@@ -305,8 +305,8 @@ def check_warning(message_regex, text):
       self.assertEqual(1, error.lineno)
       self.assertRegex(str(error), message_regex)
 
-    check_warning("Unknown pytype directive.*disalbe.*",
-                  "# pytype: disalbe=test-error")
+    check_warning("Unknown pytype directive.*disable.*",
```
rchen152, line 308: I'm pretty sure this is an intentional misspelling - the test is making sure that pytype reports an error when "disable" is spelled wrong.

rchen152 (changes requested):
Thanks! Left one comment - I think the changes to directors_test.py need to be reverted.

pytype/directors_test.py
kianmeng: @rchen152 Noted. Changes reverted. Please check again.

rchen152 (approved):
Thanks again! I'm going to merge this PR as-is and fix the lint errors when importing it into the Google codebase. (That means they'll be resolved when I re-export.)

kianmeng:
> Thanks again! I'm going to merge this PR as-is and fix the lint errors when importing it into the Google codebase. (That means they'll be resolved when I re-export.)

@rchen152 :+1: :tada:

PR #1118, opened 2022-02-07
Move Python 3.10 tests out of experimental and add 3.10 wheels

### LINKED ISSUE #1022: Support Python 3.10
Python 3.10.0 came out yesterday.

lgeiger:
Python 3.10 has been stable for a while. This PR enables testing against against 3.10 by default and adds support for building wheels on CI.

Fixes #1022

Let's see what CI things 🤞

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
         platform:
           - manylinux2014_x86_64
           - manylinux2014_aarch64
-        pyver: [cp37-cp37m, cp38-cp38, cp39-cp39]
+        pyver: [cp37-cp37m, cp38-cp38, cp39-cp39, cp310-cp310]
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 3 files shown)


rchen152:
Sadly, the reason our 3.10 tests are "experimental" isn't because of the status of Python 3.10 itself; it's because pytype's support for 3.10 isn't fully baked (many tests fail). So we aren't ready for this PR yet.

lgeiger:
Sorry about that, I didn't realise that the 3.10 experimental tests are actually ignored on CI 👍

ngam:
@rchen152 how's py310 support coming along? thoughts on enabling it in conda-forge? https://github.com/conda-forge/pytype-feedstock/pull/53

ngam:
We can override this part, but we'd like to hear your thoughts first https://github.com/google/pytype/blob/c92d839babb061b36d909bdb8481d1648cabc074/setup.cfg#L29

rchen152:
We still have 80-something 3.10 test failures, including many cases in which the pytype VM basically just blows up (crashes due to an error like wrong number of stack frames), so I can't in good conscience recommend that anyone use pytype in 3.10 yet ^^;

I'm working on basic 3.10 support in between other features, but I expect it'll be on the order of ~weeks before it's ready.

ngam:
Thanks @rchen152! We won't ship build it then 😄

Issue #1104, opened 2022-01-11
Inheritance from generic dataclass deduces wrong __init__


alaurens:
Hello,

This bug is similar to issue #628. When inheriting from a Generic dataclass, pytype fails to deduct the correct __init__ for the new class. This happens within the same file.

Here is a minimal example:

```python
import dataclasses
from typing import Generic TypeVar

K = TypeVar('K')

@dataclasses.dataclass
class BaseClass(Generic[K]):
  arg1: K

@dataclasses.dataclass
class ChildClass(BaseClass[int]):
  pass

ChildClass(arg1=3)
```

this will return the following error:
```
Invalid keyword argument arg1 to function ChildClass.__init__ [wrong-keyword-args]
         Expected: (self)
  Actually passed: (self, arg1)
```

Note however, that this code doesn't throw an error:

```python
import dataclasses
from typing import Generic TypeVar

K = TypeVar('K')

@dataclasses.dataclass
class BaseClass(Generic[K]):
  arg1: K

@dataclasses.dataclass
class ChildClass(BaseClass):
  pass

child = ChildClass(arg1=3)
child.arg1. # type: Any
```
 However in this case the type information regarding arg1 is lost, pytype no longer identifies it as an int.

This last snippet exhibits the desired behaviour, arg1 is correctly identified as an int.

```python
import dataclasses
from typing import Generic TypeVar

K = TypeVar('K')

@dataclasses.dataclass
class BaseClass(Generic[K]):
  arg1: K

@dataclasses.dataclass
class ChildClass(BaseClass[int]):
  pass

child = ChildClass(arg1=3)  # pytype: disable=wrong-keyword-args
child.arg1  # type: int
```

PR #1218, opened 2022-05-31
Fix parameterized attributes for dataclass subclasses

rgoncalvesse:
In subclasses of a Generic dataclass, the init method accepts the right
argument count, instead of expecting the argument number from its
parent's init method and raising a wrong-arg-count error.

Also add support for checking parameterized types of dataclass and raising
wrong-arg-types.

Fixes https://github.com/google/pytype/issues/1104

### DIFF
--- a/pytype/abstract/class_mixin.py
+++ b/pytype/abstract/class_mixin.py
@@ -6,6 +6,7 @@ from typing import Any, List, Optional, Type
 
 from pytype import datatypes
 from pytype.abstract import abstract_utils
+from pytype.abstract import _classes
 from pytype.abstract import function
 from pytype.abstract import mixin  # pylint: disable=unused-import
 from pytype.pytd import mro
@@ -543,6 +544,21 @@ class Class(metaclass=mixin.MixinMeta):  # pylint: disable=undefined-variable
           base_attrs.append(a)
     return base_attrs + cls_attrs
 
+  def _recompute_attrs_type_from_mro(self, all_attrs, type_params):
+    """Traverse the MRO and apply Generic type params to class attributes.
+
+    This IS REQUIRED for dataclass instances that inherits from a Generic.
+
+    Args:
+      all_attrs: All __init__ attributes of a class.
+      type_params: List of ParameterizedClass instances that will override
+        TypeVar attributes in all_attrs.
+    """
+    for typ_name, typ_obj in type_params.items():
+      for attr in all_attrs.values():
+        if typ_name == attr.typ.cls.name:
+          attr.typ = typ_obj
+
   def _get_attrs_from_mro(self, cls_attrs, metadata_key):
     """Traverse the MRO and collect base class attributes for metadata_key."""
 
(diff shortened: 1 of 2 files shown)


rgoncalvesse:
Hi, this is a hacky solution for supporting Generic and TypeVar parameters in dataclass inheritance. I have managed to get it working on similar example of https://github.com/google/pytype/issues/1104 . I will add propper tests next.

pytype/abstract/class_mixin.py
```diff
@@ -552,20 +568,31 @@ def _get_attrs_from_mro(self, cls_attrs, metadata_key):
...
+        base_cls = base_cls.base_cls
       if metadata_key in base_cls.metadata:
         sub_attrs.append([a for a in base_cls.metadata[metadata_key]
                           if a.name not in attributes_to_ignore])
     sub_attrs.append(cls_attrs)
     for attrs in sub_attrs:
       for a in attrs:
         all_attrs[a.name] = a
+
+    # type_params is only dict if it has been assigned in a dataclass context.
```
martindemello, line 593: why would type_params only be a dict for dataclasses?

pytype/abstract/class_mixin.py
```diff
@@ -552,20 +568,31 @@ def _get_attrs_from_mro(self, cls_attrs, metadata_key):
...
       if not isinstance(base_cls, Class):
         continue
       # Some third-party dataclass implementations add implicit fields that
       # should not be considered inherited attributes.
       attributes_to_ignore.update(getattr(base_cls, "IMPLICIT_FIELDS", ()))
+      # Any subclass of a Parameterized dataclass must inherit attributes from
+      # its parent's init.
+      # See https://github.com/google/pytype/issues/1104
+      if isinstance(base_cls, _classes.ParameterizedClass):
+        type_params = base_cls._formal_type_parameters
```
martindemello, line 583: use `base_cls.formal_type_parameters`, which takes care of lazily loading the attribute.

martindemello (changes requested):

pytype/abstract/class_mixin.py
```diff
@@ -579,8 +578,8 @@ def _get_attrs_from_mro(self, cls_attrs, metadata_key):
       # Any subclass of a Parameterized dataclass must inherit attributes from
       # its parent's init.
       # See https://github.com/google/pytype/issues/1104
-      if isinstance(base_cls, _classes.ParameterizedClass):
-        type_params = base_cls._formal_type_parameters
+      if abstract_utils._isinstance(base_cls, "ParameterizedClass"):
```
martindemello, line 581: can just be `if _isinstance(...)`, we have it aliased somewhere near the top of the file.

pytype/abstract/class_mixin.py
rgoncalvesse: My previous change for comparing `type_param` as a dict is not needed anymore. I had issues last week when running tests, more particularly over `chex_overlay`, then `type_param` was not a dict. Tests are okay now. Reverted in lasted commit.

martindemello:
thanks, this looks good! we'll merge it internally (with your github id) and reexport it.

PR #1228, opened 2022-06-18
Add `ParamSpecArgs` and `ParamSpecKwargs` to `typing.pytd`

AlexWaygood:
At runtime, `typing_extensions` does this with `ParamSpecArgs` and `ParamSpecKwargs`:

```python
if hasattr(typing, 'ParamSpecArgs'):
    ParamSpecArgs = typing.ParamSpecArgs
    ParamSpecKwargs = typing.ParamSpecKwargs
else:
    class ParamSpecArgs:
        # Runtime body
    class ParamSpecKwargs:
        # Runtime body
```

I'd like to do something similar in the typeshed stub, so that typeshed's stub for `typing_extensions` can match the runtime as closely as possible, i.e.:

```python
if sys.version_info >= (3, 10):
    from typing import ParamSpecArgs as ParamSpecArgs, ParamSpecKwargs as ParamSpecKwargs
else:
    class ParamSpecArgs:
        __origin__: ParamSpec
        def __init__(self, origin: ParamSpec) -> None: ...

    class ParamSpecKwargs:
        __origin__: ParamSpec
        def __init__(self, origin: ParamSpec) -> None: ...
```

However, if I try to do that at the moment, pytype crashes with a `BadDependencyError`. I believe this is due to the fact that `ParamSpecArgs` and `ParamSpecKwargs` do not currently exist in pytype's `typing.pytd`.


AlexWaygood:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

(I've now signed the CLA.)

rchen152 (approved):

rchen152:
Thanks! I'll merge this internally and re-export it.

PR #1237, opened 2022-06-24
pyupgrade --py37-plus **/*.py

cclauss:
[`pyupgrade --py37-plus **/*.py`](https://github.com/asottile/pyupgrade/blob/main/README.md)

Python f-strings are shorter, easier to read, and faster.  https://www.scivision.dev/python-f-string-speed

### DIFF
--- a/pytype/tools/merge_pyi/test_data/imports_multiple.pep484.py
+++ b/pytype/tools/merge_pyi/test_data/imports_multiple.pep484.py
@@ -1,5 +1,5 @@
 """Test multiple from-imports."""
 from typing import List, Text
 
-def f() -> List[Text]:
+def f() -> List[str]:
     return []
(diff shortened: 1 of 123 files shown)


pytype/tools/merge_pyi/test_data/imports_multiple.pep484.py
```diff
@@ -1,5 +1,5 @@
 """Test multiple from-imports."""
 from typing import List, Text
 
-def f() -> List[Text]:
+def f() -> List[str]:
```
rchen152, line 4: Could you revert this change? This is a test file anyway, and it doesn't really make sense for Text to be imported but not used.

rchen152 (changes requested):
Thanks for this PR! Just one minor comment.

pytype/tools/merge_pyi/test_data/imports_multiple.pep484.py
cclauss, line 4:
```suggestion
def f() -> List[Text]:
```

rchen152 (approved):
Thanks! I'll merge this once the checks finish running.

PR #1242, opened 2022-06-29
Windows support

wayhowang:
I tried to add windows support for pytype. It seems to work now. The package installed with `pip install -e .` works fine now. At least my needs has been satisfied. 

### Changes
* Replace delegate initializers with ordinary ones.
* Update pybind11 to v2.9.0. Previous version of pybind11 causes compiler errors on msvc ( https://github.com/pybind/pybind11/issues/3477).
* Add `tools/path_tools.py`. It provides functions with the same signatures as thoes defined in `os.path` and `glob`, but they replace the `\` in return value with `/` if pytype is running on Windows. 
* Add `tools/tempfile.py`. Files created by NamedTemporaryFile cannot be opened twice in the same process on Windows if by default. `tools/tempfile.py` provides an alternative `NamedTemporaryFile ` implementation for Windows. See https://stackoverflow.com/questions/23212435/permission-denied-to-write-to-my-temporary-file for details.
* Edit several `CMakeLists.txt`s

### Current Issue
I ran `pip install -e . && python build_scripts/run_tests.py` to test it. It passed every test cases before following text showed in the command line. It seems there are some problems with the build system. (But building with `pip install -e .` works fine.)
```
>>> Found Ninja target failures (includes test failures):
    - pytype/typegraph/cfg.so pytype/typegraph/cfg.lib
>>> FAILED: Ninja command 'ninja -k 100000 test_all'.
>>>         Run it in the 'out' directory to reproduce.
>>>         Full Ninja output is available in 'C:\Users\wweih\Documents\pytype-modified\pytype\out\ninja.log'.
>>>         Failing test modules (if any) will be reported below.
```
[ninja.log](https://github.com/google/pytype/files/9012640/ninja.log)

### Help Wanted
* I am not expertize in building systems. Are there any advice about the above issue?
* Which tool should I use in order to format the source code?
* `/` and `\` are both valid seperators on Windows. It caused a lot of problems for pytype. I tried to solve this by replacing all `\`  with `/` on windows. Is this an acceptable solution?
* Is there any automated test pipline to test my commits on Linux and macOS? Or should I setup an Linux system to test it on myself?


wayhowang:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/google/pytype/pull/1242/checks?check_run_id=7118009489) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

I have signed now.. What should I do after that？

rchen152:
Wow, thanks for doing this!

> I am not expertize in building systems. Are there any advice about the above issue?

Odd error, I haven't seen that before! The relevant section of the log seems to be:

```
FAILED: pytype/typegraph/cfg.so pytype/typegraph/cfg.lib 

cmd.exe /C "cd . && "C:\Program Files\CMake\bin\cmake.exe" -E vs_link_dll --intdir=pytype\typegraph\CMakeFiles\pytype.typegraph.cfg.dir --rc=C:\PROGRA~2\WI3CF2~1\10\bin\100190~1.0\x64\rc.exe --mt=C:\PROGRA~2\WI3CF2~1\10\bin\100190~1.0\x64\mt.exe --manifests  -- C:\PROGRA~2\MICROS~2\2019\COMMUN~1\VC\Tools\MSVC\1429~1.301\bin\Hostx64\x64\link.exe /nologo pytype\typegraph\CMakeFiles\pytype.typegraph.cfg.dir\cfg.cc.obj pytype\typegraph\CMakeFiles\pytype.typegraph.cfg.dir\cfg_logging.cc.obj pytype\typegraph\CMakeFiles\pytype.typegraph.cfg.dir\pylogging.cc.obj pytype\typegraph\CMakeFiles\pytype.typegraph.cfg.dir\reachable.cc.obj pytype\typegraph\CMakeFiles\pytype.typegraph.cfg.dir\solver.cc.obj pytype\typegraph\CMakeFiles\pytype.typegraph.cfg.dir\typegraph.cc.obj  /out:pytype\typegraph\cfg.so /implib:pytype\typegraph\cfg.lib /pdb:pytype\typegraph\cfg.pdb /dll /version:0.0 /machine:x64 /debug /INCREMENTAL  C:\Users\wweih\AppData\Local\Programs\Python\Python310\libs\python310.lib  kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib  && cd ."

LINK Pass 1: command "C:\PROGRA~2\MICROS~2\2019\COMMUN~1\VC\Tools\MSVC\1429~1.301\bin\Hostx64\x64\link.exe /nologo pytype\typegraph\CMakeFiles\pytype.typegraph.cfg.dir\cfg.cc.obj pytype\typegraph\CMakeFiles\pytype.typegraph.cfg.dir\cfg_logging.cc.obj pytype\typegraph\CMakeFiles\pytype.typegraph.cfg.dir\pylogging.cc.obj pytype\typegraph\CMakeFiles\pytype.typegraph.cfg.dir\reachable.cc.obj pytype\typegraph\CMakeFiles\pytype.typegraph.cfg.dir\solver.cc.obj pytype\typegraph\CMakeFiles\pytype.typegraph.cfg.dir\typegraph.cc.obj /out:pytype\typegraph\cfg.so /implib:pytype\typegraph\cfg.lib /pdb:pytype\typegraph\cfg.pdb /dll /version:0.0 /machine:x64 /debug /INCREMENTAL C:\Users\wweih\AppData\Local\Programs\Python\Python310\libs\python310.lib kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib /MANIFEST /MANIFESTFILE:pytype\typegraph\CMakeFiles\pytype.typegraph.cfg.dir/intermediate.manifest pytype\typegraph\CMakeFiles\pytype.typegraph.cfg.dir/manifest.res" failed (exit code 1104) with the following output:
LINK : fatal error LNK1104: cannot open file 'python310_d.lib'
```

I found https://github.com/pybind/pybind11/issues/3403, which looks like a very similar issue (same error message, and we use pybind11). It would appear that the problem appears when Windows Python *with debug libraries* is installed. Maybe try not including the debug libraries and see if that helps?

> Which tool should I use in order to format the source code?

We don't use an autoformatter, but if you'd like to, I believe pyformat (https://pypi.org/project/pyformat/) will format your code in a way that is compatible with the linter check.

> / and \ are both valid seperators on Windows. It caused a lot of problems for pytype. I tried to solve this by replacing all \ with / on windows. Is this an acceptable solution?

The one potential issue I see is that if we want to reconstruct the original path, we won't be able to if `\` and `/` are indistinguishable. But if recovering the original path isn't necessary, then this approach sounds fine to me.

> Is there any automated test pipline to test my commits on Linux and macOS? Or should I setup an Linux system to test it on myself?

I believe you can run our tests via GitHub Actions in your forked pytype repository. If you go to the "Actions" tab of the repository, you should be able to enable the workflows, then select a branch to run them on. You'll probably mostly want the "CI" workflow, which does linting, testing, and type-checking.

> I have signed now.. What should I do then？Wait for a repository owner to approve me to run workflows again? Sorry that I am not familar with the opensource community.

It looks like the CLA check is still failing, actually. In the failure details, it says that "PR Opener: @wayhowang" has signed the CLA, but "Author: @yslking <ww****o​@outlook.com>" has not. It looks like you might have different email addresses associated with your commit messages and with your GitHub account? I think if you change one of them so that the two are using the same address, then the check should be happy.

wayhowang:
> It looks like the CLA check is still failing, actually. In the failure details, it says that "PR Opener: @wayhowang" has signed the CLA, but "Author: @yslking <ww****o​@outlook.com>" has not. It looks like you might have different email addresses associated with your commit messages and with your GitHub account? I think if you change one of them so that the two are using the same address, then the check should be happy.

I will close this PR and open another later.

> I found https://github.com/pybind/pybind11/issues/3403, which looks like a very similar issue (same error message, and we use pybind11). It would appear that the problem appears when Windows Python with debug libraries is installed. Maybe try not including the debug libraries and see if that helps?

Thanks a lot. Exactly what I need. 

> The one potential issue I see is that if we want to reconstruct the original path, we won't be able to if \ and / are indistinguishable. But if recovering the original path isn't necessary, then this approach sounds fine to me.

I choosed this solution because it was easy for me to do this modification. That is, replace `os.path.` with `path_tools`., add import statement, and add dependencies in `CMakeLists.txt`s. All these works could be done using scripts. However, in addition to the potential issue you mentioned, I found things will be difficult to deal with if third party libraries also return paths. In `pytype`, `importlab` will return a path separated by `os.path.sep`, and those paths are deeply hidden in data structures  like nx.Graph, making it hard to replace `os.path.sep` with `/`.

As far as I have known, a path in windows can be alternated in three ways while keeping its semantic unchanged. 
1. Change the case of any letter;
2. Change separator bidirectionally (in most cases);
3. Change between long path (e.g., `C:\Program Files`) or short path (e.g., `C:\PROGA~1`) 

The most comprehensive solution I can think up is to use a special `Path` class, and rewrite its `__hash__` and `__eq__`. But I believe there will be two major shortcomings.
1. It's too time consuming to find every piece of codes that involves path handling.
2. New bugs might (perhaps must)  be introduced to `pytype` on Linux and macOS temporarily.

Currently I choose to use `os.path.sep` as the only acceptable path separator on Windows. `pytype` seems work better when I use `os.path.sep` instead of `/` as the separator on Windows.  (It passed `build_scripts/run_tests.py` except one case now). And I have replaced driver code returned by `os.path.` functions with upper case ones. Still, there will be problems with short and long path, lower and upper case letter (sometimes python returns driver code in lowercase), and problems when do path reconstruction. And users have to use `\\` as the seperator in Windows when passing commond line arguments and write config files, though `/` should be supported (btw, python installer itself don't support `/`-seperated path). 

I am still a little confused about "reconstruct the original path" now. What is the purpose of reconstruction? If the reconstructed paths are used to print paths in error reports or used as the target path for I/O, then I guess there should not be too much problems to change the path's representation while keeping their semantic.


[This](https://github.com/wayhowang/pytype/blob/b7c40eeb2d09031b18f2ffe22207b5b7ca267868/pytype/tests/test_match2.py#L290) is the only test cases that was not passed. I am confused about its behavior. Details will be posted in another comment.

wayhowang:
> [This](https://github.com/wayhowang/pytype/blob/b7c40eeb2d09031b18f2ffe22207b5b7ca267868/pytype/tests/test_match2.py#L290) is the only test cases that was not passed. I am confused about its behavior. Details will be posted in another comment.

The checked code is
```
import collections
import six
from typing import Dict

def f() -> Dict[str, bytes]:
    d = collections.defaultdict(list)
    for _ in range(10):
        subdict = {}  # type: Dict[str, str]
        k = subdict.get('k')
        if not k:
            continue
        d[k].append(b'')
    return {k: b', '.join(v) for k, v in six.iteritems(d)}
```

The error reported on Windows:
```
...... bad return type [bad-return-type]
           Expected: Dict[str, bytes]
  Actually returned: Dict[None, bytes]
```

I did two exprements on it.

### Exp.1st
I tried to print out the type of expressions using the following script:
```
import pytype
import pytype.io
import pytype.config
import pytype.pytd.pytd as pytd
import pytype.tools.traces
import pytype.tools.traces.traces
import pytype.pytd.pytd_utils
import pytype.tools.annotate_ast.annotate_ast as aast
import ast
from typing import *


source = """

from collections import *
import six
from typing import Dict

def f() -> Dict[str, bytes]:
    d = defaultdict(list)
    for _ in range(10):
        subdict = {'k': 'v'}  # type: Dict[str, str]
        k = subdict.get('k')
        if not k:
            continue    
        d[k].append(b'')
    d_out = d
    out = {k: b', '.join(v) for k, v in six.iteritems(d)}
    return out

"""

config = pytype.config.Options.create()   
module = aast.annotate_source(source, ast, pytype.config.Options.create())
for node in ast.walk(module):
    resolved_type = getattr(node, 'resolved_type', None)
    if not resolved_type:
        continue
    if not isinstance(resolved_type, pytd.Type):
        continue
    print(f"## {node.resolved_annotation:60} {repr(ast.unparse(node))}")
```

* On Windows, the type of `d_out` is `collections.defaultdict[Optional[str], Any]`.
* On Linux (with the unchanged pytype, a1ec5211c), the type of `d_out` is `collections.defaultdict[None, Any]`

The reported type on windows is not accurate enough but I think the one on Linux in wrong. 

### Exp. 2nd

I test the following code on Linux with `pytype <filename>`
```
import collections
import six
from typing import Dict

def f() -> Dict[str, bytes]:
    d = collections.defaultdict(list)
    for _ in range(10):
        subdict = {}  # type: Dict[str, str]
        k = subdict.get('k')
        if not k:
            continue
        not_non_k = k               # ADDED LINE
        d[not_non_k].append(b'')    # ADDED LINE
        d[k].append(b'')
    return {k: b', '.join(v) for k, v in six.iteritems(d)}
```

But it reported 
```
File "/home/wayhoo/test_pytype/test3.py", line 15, in f: bad return type [bad-return-type]
           Expected: Dict[str, bytes]
  Actually returned: Dict[None, bytes]
```

Are these expected behaviors ?

martindemello:
that's really weird; i can't understand it either. could you try a couple of experiments on your branch:

1. see if the tests pass using `items` rather than `six.iteritems`
2. see if `dis.dis` shows any differences in the code on windows and linux (since python is bytecode based, if there are compilation differences some test might be affected)

wayhowang:
Thanks for your suggestion. 
> On Linux (with the unchanged pytype, https://github.com/google/pytype/commit/a1ec5211c7070ec8d051253f15e14f451ed747f3), the type of d_out is collections.defaultdict[None, Any]

Is the above thing expected? I think `d_out` should be of `collections.defaultdict[str, bytes]`. I ran this test on `google/pytype:main`

> see if the tests pass using items rather than six.iteritems

Nothing changed. But if i change `defaultdict` to `dict`, then it pass test on Windows.

<table>
<thead>
  <tr>
    <th>Test Case</th>
    <th>main branch of google/pytype</th>
    <th>my branch (Linux)</th>
    <th>my branch (Windows 10)</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td>

```
import collections
import six
from typing import Dict
def f() -> Dict[str, bytes]:
  d = collections.defaultdict(list)
  for _ in range(10):
    subdict = {}  # type: Dict[str, str]
    k = subdict.get('k')
    if not k:
      continue
    d[k].append(b'')
  return {k: b', '.join(v) for k, v in six.iteritems(d)}
```

</td>
    <td>Success: no errors found</td>
    <td>Success: no errors found</td>
    <td>bad return type [bad-return-type]
           Expected: Dict[str, bytes]
  Actually returned: Dict[None, bytes]</td>
  </tr>
  <tr>
    <td>

```
import collections
import six
from typing import Dict
def f() -> Dict[str, bytes]:
  d = collections.defaultdict(list)
  for _ in range(10):
    subdict = {}  # type: Dict[str, str]
    k = subdict.get('k')
    if not k:
      continue
    d[k].append(b'')
  return {k: b', '.join(v) for k, v in d.items()}   # CHANGED
```

</td>
    <td>Success: no errors found</td>
    <td>Success: no errors found</td>
    <td>bad return type [bad-return-type]
           Expected: Dict[str, bytes]
  Actually returned: Dict[None, bytes]</td>
  </tr>
  <tr>
    <td>

```
import collections
import six
from typing import Dict
def f() -> Dict[str, bytes]:
  d = dict()
  for _ in range(10):
    subdict = {}  # type: Dict[str, str]   # CHANGED
    k = subdict.get('k')
    if not k:
      continue
    if not k in d:      # CHANGED
      d[k] = [b'']
    else:
      d[k].append(b'')
  return {k: b', '.join(v) for k, v in d.items()}
```

</td>
    <td>Success: no errors found</td>
    <td>Success: no errors found</td>
    <td>Success: no errors found</td>
  </tr>
</tbody>
</table>

> see if dis.dis shows any differences in the code on windows and linux (since python is bytecode based, if there are compilation differences some test might be affected)



I ran this code on both OS. No substantial difference. The only difference is the path of the source code.
```
import collections
from typing import Dict
def f() -> Dict[str, bytes]:
  d = collections.defaultdict(list)
  for _ in range(10):
    subdict = {}  # type: Dict[str, str]
    k = subdict.get('k')
    if not k:
      continue
    d[k].append(b'')
  return {k: b', '.join(v) for k, v in d.items()}

import dis
print(dis.dis(f))
```

martindemello:
okay, i'm mystified too, but in the interests of not blocking this PR you can add a `@skipIf(sys.platform  == "win32")` decorator to that one test.

martindemello:
btw a simpler way to see what pytpe thinks the type of something is is to add a `reveal_type(expression)` in your code. pytype will emit a type error with the type of the expression.

build_scripts/ci_script.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 """A simple script to run CI build steps."""
 
+
```
martindemello, line 4: revert all the changes that merely add a newline to a file, in the interests of not touching unnecessary files in an already large change

pytype/tools/CMakeLists.txt
```diff
@@ -9,6 +9,8 @@ py_library(
     .environment
     .runner
     .tool_utils
+    .path_tools
```
martindemello, line 12: the `tools` directory is for standalone tools that build on top of pytype. pytype itself should not depend on anything in this subtree. also for naming consistency, call the module `path_utils` rather than `path_tools`

if you would like to group the win32 wrappers together you could create a new `pytype/platform_utils` directory, otherwise just move both these modules to the top level `pytype/`

pytype/tools/path_tools.py
```diff
@@ -0,0 +1,47 @@
+import sys
```
martindemello, line 1: add a module docstring

pytype/tools/tempfile.py
```diff
@@ -0,0 +1,24 @@
+import tempfile
```
martindemello, line 1: add a module docstring here too

martindemello (changes requested):

pytype/typegraph/cfg.cc
```diff
@@ -499,51 +499,52 @@ PyDoc_STRVAR(
     "stay in memory.");
 
 PyTypeObject PyProgram = {
-  PyVarObject_HEAD_INIT(&PyType_Type, 0) tp_name : "Program",
-  tp_basicsize : sizeof(PyProgramObj),
-  tp_itemsize : 0,
-  tp_dealloc : ProgramDealloc,
+  PyVarObject_HEAD_INIT(&PyType_Type, 0)
```
martindemello, line 502: msvc added support for designated initializers in 2019; what would the impact of not making this change be? (i'm fine  with leaving it in if a significant number of users would be impacted, or if it is onerous to upgrade msvc for people who have an older version, but it makes the code more fragile)

wayhowang: Mixing old-style syntax and designated initializer together is [not supported](https://en.cppreference.com/w/cpp/language/aggregate_initialization#:~:text=Note%3A%20out%2Dof%2Dorder%20designated%20initialization%2C%20nested%20designated%20initialization%2C%20mixing%20of%20designated%20initializers%20and%20regular%20initializers%2C%20and%20designated%20initialization%20of%20arrays%20are%20all%20supported%20in%20the%20C%20programming%20language%2C%20but%20are%20not%20allowed%20in%20C%2B%2B.) by neither c++20 standard nor msvc. I guess there are some parameters written in old-style syntax in `PyVarObject_HEAD_INIT`.   Original code cause compile-time errors in VS19/22 even if language standard is set to c++20 or latest.

wayhowang: Sure. It will be excellent if designated initializers can be kept. I have restored this file.

martindemello (approved):
looks good! we'll merge it internally and reexport it to github

martindemello:
this should be fully merged now, and we can see what to do about the designated initializers.

PR #1245, opened 2022-06-30
modified requirements.txt by removing unused stuff

prasadjivane:
Removed unused data classes and typed_ast also modified versions.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@
 attrs>=21.4.0
 importlab>=0.7
 libcst>=0.4.5
-networkx>=2.6.2
+networkx<2.8.4
 ninja>=1.10.0.post2
 pybind11>=2.9.2
 pydot>=1.4.2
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -2,14 +2,12 @@
 # Make sure you also install the non-Python dependencies described in
 # https://github.com/google/pytype/blob/master/CONTRIBUTING.md#pytype-dependencies.
 attrs>=21.4.0
-dataclasses; python_version < '3.7'
 importlab>=0.7
-libcst
-networkx<2.8.4
+libcst>=0.4.5
+networkx>=2.6.2
```
rchen152, line 7: I'm pretty sure we need to keep the <2.8.4; last I checked, using networkx 2.8.4 caused some odd test failure.

rchen152 (changes requested):
Thanks for this change! Could you also update the dependencies in setup.cfg?

requirements.txt
rchen152: So please add <2.8.4 back. >=2.6.2 will not prevent version 2.8.4 from being installed.

rchen152 (approved):
Thanks! I added the typed_ast line back, since we do use typed_ast in 3.7. (It gets pulled in anyway as an indirect dependency, but it's better to declare what we use.) Other than that, this looks great now!

PR #1276, opened 2022-08-22
Change Environment construction to prefer typeshed

Penguinwizzard:
Currently, typeshed only gets hit if nothing else does at an import parsing level in importlab. This dramatically increases the amount of time spent on standard library components in some situations. If we instead check typeshed first, we can aggressively take advantage of pre-created pyi files for typechecking performance.


rchen152:
Unfortunately, I don't think this change is correct. We check typeshed after pythonpath because Python allows local modules to shadow stdlib ones.

PR #1335, opened 2022-12-24
Add Python 3.11 production release to ci testing

cclauss:


rchen152 (approved):
Thanks!

cclauss:
I believe that if we remove the `experimental: true` flag that the Python 3.11 tests will fail.

PR #1351, opened 2023-01-12
Improve syntax highlighting of `.pytd` files on GitHub

AlexWaygood:
This PR teaches GitHub that `.pytd` files use Python syntax. This means that GitHub will apply much better syntax highlighting to `.pytd` files in pytype's source code.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Tell GitHub to use Python syntax highlighting for `.pytd` files
+*.pytd linguist-language=python


AlexWaygood:
Using a `.gitattributes` file for this purpose is documented here: https://github.com/github/linguist/blob/master/docs/overrides.md#using-gitattributes

rchen152 (approved):

rchen152:
Thank you!

Issue #1376, opened 2023-02-20
pytype not using system ninja, even if it's there


PythonCHB:
pytype using ninja (honestly, I haven't looked into how or why, but I'm sure you all know :-) (and I may be getting some of this wrong, but ...)

However, ninja is not a python package, so it has to be installed somehow.

For Python / pip users, a "ninja" package is on PyPi that is basically a wrapper around ninja so that it can be installed by pip. Nifty.

However, at least as of commit: 1a6d0b3481c7 -- pytype seems to always try to call ninja via the python wrapper, which fails if the python wrapper isn't there, even though the ninja executable is there.

This is an issue for the conda-forge pytype package, which has been broken for some time (still trying to figure out when it last worked). conda-forge has a ninja package that installs the ninja executable, which works fine itself, but not with the current pytype package.

I imagine it would be an issue for any other system that can install ninja itself -- e.g. Linux distros.

This is discussed in the conda-forge "feedstock" repo here:

https://github.com/conda-forge/pytype-feedstock/issues/97

eli-schwartz:
I would suggest checking for both `shutil.which('ninja')` and `from ninja import ninja` to see which one is available.

PythonCHB:
> `from ninja import ninja`

Make sure to wrap that in a:
```
try:
    from ninja import ninja
except ImportError:
    ...
```

PR #1378, opened 2023-02-22
use ninja importability to determine python-ninja usability

ngam:
Fixes https://github.com/google/pytype/issues/1376

There doesn't seem to be a need to force people to use python-ninja when the ninja binary is available. The logic is better now such that if python-ninja isn't available (importable), the command can fall back to ninja the binary.

xref:
- https://github.com/conda-forge/pytype-feedstock/pull/98
- https://github.com/conda-forge/staged-recipes/pull/19098

thanks to:
- @eli-schwartz
- @PythonCHB

### DIFF
--- a/pytype/tools/analyze_project/pytype_runner.py
+++ b/pytype/tools/analyze_project/pytype_runner.py
@@ -48,7 +48,7 @@ def _get_executable(binary, module=None):
               'pytype-single')
       ])
   importable = importlib.util.find_spec(module or binary)
-  if sys.executable is not None and importable is not None:
+  if sys.executable is not None and importable:
     return [sys.executable, '-m', module or binary]
   else:
     return [binary]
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Also xref: #1376 (should be marked as closing that ticket).

pytype/tools/analyze_project/pytype_runner.py
```diff
@@ -46,7 +47,8 @@ def _get_executable(binary, module=None):
               path_utils.abspath(path_utils.dirname(custom_bin)),
               'pytype-single')
       ])
-  if sys.executable is not None:
+  importable = importlib.util.find_spec(module or binary)
+  if sys.executable is not None and importable is not None:
```
martindemello, line 51: this can be just `... and importable` since `find_spec` does not return anything else that evaluates to `False`

martindemello:
thanks, that looks like a great fix! sorry for all the conda issues.

pytype/tools/analyze_project/pytype_runner.py
martindemello: i prefer the two-liner, the variable name `importable` does add a nice bit of readability.

martindemello:
looks good, i'll merge it internally (retaining you as author) and push it out as part of the next pytype release

PR #1386, opened 2023-03-01
Make it possible to run pytype on PyPy.

manueljacob:


pytype/metrics.py
```diff
@@ -336,6 +340,8 @@ class Snapshot(Metric):
 
   def __init__(self, name, enabled=False, groupby="lineno",
                nframes=1, count=10):
+    if enabled and tracemalloc is None:
+      raise Exception("tracemalloc module couldn't be imported")
```
rchen152, line 344: Could you raise a more specific exception type? (Maybe ImportError?)

manueljacob: At this point, I think that raising `ImportError` is not perfect because the import happened before (and was the cause of the error), but here we want to notify that the snapshot metric doesn’t work, which is a different kind of error. But I agree that raising a very general exception type is also not very good. Unless you have another idea, I would be fine with both leaving and changing it.

rchen152: How about `ValueError` or `RuntimeError`? I'm not too fussed about which exception class you pick, as long as it makes some sense, but I'd strongly prefer not to leave it as an entirely uninformative `Exception`.

manueljacob: Thanks for the suggestion. I changed it.

rchen152 (approved):

rchen152:
Merged as https://github.com/google/pytype/pull/1388/commits/28707444b9123174d458cd3571d444d379297112.

Issue #1415, opened 2023-04-25
Pytype won't accept a single file without `.py` extension as input


redat00:
Hello ! 

I'm working on a small script. This script, as it is an executable, does not have a `.py` extension, and is not in a directory.  It's the file called `s3`.

```bash
.
├── COPYING
├── poetry.lock
├── pyproject.toml
├── README.md
└── s3
```

I wan't to run `pytype` on it, but I'm not able to since I have this error message : 
```
# poetry run pytype s3
usage: pytype [options] input [input ...]
pytype: error: Need an input.
```

However, if I just rename the file from `s3` to `s3.py` I don't have this problem anymore. 
Is it something that should occur, or is there something I'm missing ? 

Thanks in advance for any help !

rchen152:
Thank you @trevor-james-nangosha for fixing this! Next week's release will contain the fix.

PR #1419, opened 2023-04-29
Added support for python scripts. #1415

trevor-james-nangosha:
Regarding issue #1415, I have added support for Python scripts by checking for a shebang at the start of the file with the path to the Python executable.

One thing to note with this approach is that it will check all files in the directory/those that you provide to see if they are python executable scripts. That is, even if you have a file called `main.html`, but it has a python shebang line at the top and has python code, then it will be type checked.

This works for me, but I am also looking forward to any opinions that could provide more robust solutions.


rchen152:
Sorry for the slow response! This looks like it would open and read every file in a directory and subdirectories to find Python scripts? That seems like it would be too slow to be practical, unfortunately.

What do you think of only allowing directly specified files to have non-.py extensions? That is, if you run `pytype my_script`, then `my_script` should be analyzed, but pytype will still only search for `.py` files when looking for dependencies of `my_script`.

trevor-james-nangosha:
@rchen152. That makes sense. I'll get on it right away.

trevor-james-nangosha:
@rchen152, have added another commit to only check for python scripts when a non-`.py` file is specified.

pytype/file_utils.py
```diff
@@ -136,3 +140,12 @@ def replace_separator(path: str):
...
     return path
+  
+def is_file_script(filename, dir=None):
+  # This is for python files that do not have the .py extension
+  # of course we assume that they start with a shebang
+  file_path = expand_path(filename, dir)
+  if path_utils.isfile(file_path):
+    with open(file_path, 'r') as file:
+      line = file.readline().rstrip().lower()
+      return line.startswith('#!') and (line.endswith('python') or line.endswith('python3'))
```
rchen152, line 151: I'd recommend using a regex here. Something like `return re.fullmatch(r'#!.+python3?', line) is not None` should work.

pytype/file_utils.py
```diff
@@ -114,10 +114,14 @@ def expand_source_files(filenames, cwd=None):
   out = []
   for f in expand_globpaths(filenames.split(), cwd):
     if path_utils.isdir(f):
-      # If we have a directory, collect all the .py files within it.
-      out += recursive_glob(path_utils.join(f, "**", "*.py"))
+      # If we have a directory, collect all the .py files within it....
+      out  += recursive_glob(path_utils.join(f, "**", "*.py"))
```
rchen152, line 118:
```suggestion
      out += recursive_glob(path_utils.join(f, "**", "*.py"))
```

pytype/file_utils_test.py
```diff
@@ -107,9 +107,15 @@ class TestExpandSourceFiles(unittest.TestCase):
...
 
+  SCRIPT_CODE = """
+  #!/usr/bin/python3
+  print("hello, world")
+  """
+
   def _test_expand(self, string):
     with test_utils.Tempdir() as d:
       fs = [d.create_file(f) for f in self.FILES]
+      fs += [d.create_file("my_script", self.SCRIPT_CODE)]
```
rchen152, line 118: What is this line accomplishing? The test still asserts that expand_source_files only finds the .py files.

rchen152 (changes requested):
Thanks! A few comments.

pytype/file_utils_test.py
trevor-james-nangosha: Previously when I was checking every file in the directory to check if it was a script, this line would have made sense to make sure we don't miss a script file. Now that we only check for a file that is passed to `pytype`, this line  looks redundant.

I will remove it. Thanks for pointing that out.

rchen152 (changes requested):
Looking good! There are a couple of lint errors: https://github.com/google/pytype/actions/runs/4945740447/jobs/8861864375

- There's extra whitespace on line 144.
- The parameter name `dir` on line 145 conflicts with the `dir` builtin.

Once those are fixed, I think this will be ready to merge.

trevor-james-nangosha:
@rchen152, Done!!!!!!!!

PR #1454, opened 2023-06-30
use old libcst for python <= 3.9

martindemello:


rchen152:
This doesn't appear to be needed anymore, as tests pass after https://github.com/google/pytype/pull/1456.

PR #1465, opened 2023-07-13
copy over umarshal.py from cpython

martindemello:


rchen152:
I think it would be helpful to track which version of cpython this umarshal.py file is from. Maybe it could go in a version subdirectory? That would also make it easier to have multiple umarshal.py files for different versions in the future.

martindemello:
i thought of that but so far the file seems to be stable from 3.11-HEAD. i'm planning on moving this to its own github project and importing that into third_party and pypi at some point, so i just did the simplest possible thing in here for now.

gpshead:
Agreed on tracking the version.  have it live in a v3_11 and so on subpackage perhaps?

If upstream conversations about making an actual version specific stdlib API of this similar how the `dis` and `ast` modules are very version specific go in our favor, we won't wind up with too many versions - but it is good to be prepared.

gpshead:
oh, having it be its own github and PyPI project also makes sense.

rchen152:
Yeah, that makes sense!

rchen152 (approved):

PR #1582, opened 2024-02-10
minor clean-up to remove mentions of pytype.single

ngam:
In `_get_executable('pytype-single', 'pytype.single')`, the module part (`pytype.single`) wasn't used in the logic above (see below). So, to avoid confusion, just delete it altogether.

```py
  if binary == 'pytype-single':
    custom_bin = path_utils.join('out', 'bin', 'pytype')
    if sys.argv[0] == custom_bin:
      # The Travis type-check step uses custom binaries in pytype/out/bin/.
      return (([] if sys.platform != 'win32' else [sys.executable]) + [
          path_utils.join(
              path_utils.abspath(path_utils.dirname(custom_bin)),
              'pytype-single')
      ])
```

### DIFF
--- a/pytype/tools/analyze_project/pytype_runner.py
+++ b/pytype/tools/analyze_project/pytype_runner.py
@@ -52,7 +52,7 @@ def _get_executable(binary, module=None):
     return [sys.executable, '-m', module or binary]
   else:
     return [binary]
-PYTYPE_SINGLE = _get_executable('pytype-single', 'pytype.single')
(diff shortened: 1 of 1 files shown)


rchen152:
Thanks for the PR, but I don't think this is the right fix, since the module name is used in some cases. (The `return` statement is nested inside a second `if`, so it's possible to exit the `if binary == 'pytype-single':` block and enter the code where `module` is used.)

Issue #499, opened 2020-01-25
Error of paths with spaces (Linux)


staticdev:
I am using latest version of pytype with a folder with space and get the error:

```
$ pwd
/home/user/workspace/folder/subfolder/2019/test/Foregone Solution

$ pytype solution.py
Computing dependencies
Analyzing 1 sources with 0 local dependencies
ninja: Entering directory `/home/user/workspace/folder/subfolder/2019/test/Foregone Solution/.pytype'
ninja: error: dependency cycle: /home/user/workspace/folder/subfolder/2019/test/Foregone -> /home/user/workspace/folder/subfolder/2019/test/Foregone

```
If I rename the folder to `Foregone Solution` to `ForegoneSolution`

```
$ pytype solution.py
Computing dependencies
Analyzing 1 sources with 0 local dependencies
ninja: Entering directory `/home/user/workspace/folder/subfolder/2019/test/ForegoneSolution/.pytype'
[1/1] check solution
Success: no errors found
```

rchen152:
Ha, that's not the error I would've expected (dependency cycle?), but I'm not surprised that we can't handle paths with spaces. I don't think that's tested anywhere. Thanks for the report!

alewis:
A consequence is that Pytype cannot straightforwardly operate upon files store in a Google Drive (since the directory structure by default takes a form such as ~/'Google Drive File Stream'/'My Drive'/...)

ibozhilov:
I am facing the same issue, but with different error thrown:

```
ninja: Entering directory `/var/lib/jenkins/jobs/ic-iom Build and Release/workspace/.pytype'
ninja: error: build.ninja:13: multiple rules generate /var/lib/jenkins/jobs/ic-iom [-w dupbuild=err]
```

What's interesting is that ninja enters the right directory, but then obviously it does not escape the whitespaces in the path.

wyattscarpenter:
I am also encountering this error, with the "multiple rules generate" error message.

PR #1608, opened 2024-04-02
Update pytype_runner.py - change escape_ninja_path to escape all the ninja path escape-needing characters

wyattscarpenter:
This fixes https://github.com/google/pytype/issues/499

Unfortunately, I was not able to get pytype running on my machine, so this commit is just based on what seems like the right idea to me, and hopefully it can be verified to be correct by a maintainer and then merged.

### DIFF
--- a/pytype/tools/analyze_project/pytype_runner.py
+++ b/pytype/tools/analyze_project/pytype_runner.py
@@ -4,6 +4,7 @@ import collections
 import importlib
 import itertools
 import logging
+import re
 import subprocess
 import sys
 from typing import Iterable, Sequence, Tuple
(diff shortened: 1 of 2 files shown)


pytype/tools/analyze_project/pytype_runner.py
```diff
@@ -146,19 +146,31 @@ def _module_to_output_path(mod):
...
+  Escape colon, dollar sign, space, and new line, for ninja
+  (as described in https://ninja-build.org/manual.html#ref_lexer).
+  This function should only ever be used once, called on a path string to turn
+  it into a ninja path string. (If you call it on a ninja path string, it will
+  render the variables inert.)
+
+  Args:
+    path: The path.
+  """
+  new_path = ''
```
rchen152, line 160: Now that there are multiple special characters involved, I'd prefer to use `re.sub` to make the code more concise. This seems to work:
```
return re.sub(r'(?P<char>[\n :$])', r'$\g<char>', path)
```
(An `import re` line is also needed at the top of the file, of course.)

pytype/tools/analyze_project/pytype_runner.py
```diff
@@ -338,7 +350,7 @@ def write_build_statement(self, module, action, deps, imports, suffix):
     logging.info('%s %s\n  imports: %s\n  deps: %s\n  output: %s',
                  action, module.name, imports, deps, output)
     if deps:
-      deps = ' | ' +  escape_ninja_path(' '.join(deps))
+      deps = ' | ' + ' '.join([escape_ninja_path(dep) for dep in deps])
```
rchen152, line 353:
```suggestion
      deps = ' | ' + ' '.join(escape_ninja_path(dep) for dep in deps)
```

rchen152 (changes requested):
Thanks! A couple suggestions to make the code more concise.

pytype/tools/analyze_project/pytype_runner.py
wyattscarpenter: `re.sub(r'([\n :$])', r'$\1', path)` also works equivalently, and reads slightly simpler, if you prefer that.

rchen152 (approved):

Issue #1154, opened 2022-03-20
Please support Windows.


mrolle45:
**My tale of woe**
I have a Windows PC, and I wanted to run pytype on my code.  I had to install WSL and then load several needed packages.  Not only was this a big pain, but when pytype gave me some bad results, I tried debugging it.
I'm not familiar with the latest IDEs on linux, nor did I want to spend the effort of learning to debug Python code visually (and load even more packages).
So I resorted to inserting lots of print() calls in the code until I figured out what the problem was.
About two days spent on this effort.
I managed to get good pyi outputs for most, but not all, of my source code.  I then tried running merge_pyi on the good ones, only to find out that merge_pyi crashed.
On the hunch that at least merge_pyi didn't do anything weird, I installed pytype on my Windows Python, and used Visual Studio to debug it.  It worked without problems.  I was able to find and fix the bug, and file an issue about it.  See #1153.

**The pitch**
I presume that there can't be too much involved in getting the entire pytype package to run correctly on Windows.  After all, Python is mostly portable.
It would be a great assistance to Windows Python users if they could run pytype directly.
Thank you.

martindemello:
we would definitely love to have better support for windows, the problem is that there are only three pytype devs, and we all use linux. we are very open to pull requests for windows and macos support, but we are unable to do the work ourselves.

IIHERO4:
> we would definitely love to have better support for windows, the problem is that there are only three pytype devs, and we all use linux. we are very open to pull requests for windows and macos support, but we are unable to do the work ourselves.

Hey, i was just wondering what would be necessary in order to support windows.

martindemello:
@IIHERO4 just an iterative process of running pytype, finding issues, investigating and fixing them - i'm afraid we don't have a proper roadmap because we don't even know what all the issues are.

the current blocker is #90 - it's possible that everything will just work after that is fixed, but it's also possible that other issues are hiding behind it.

HexDecimal:
The designated initializers like the ones here seem to be non-standard and break MSVC.
https://github.com/google/pytype/blob/efc1fd95170e0569fa1821d960108017f072cc17/pytype/typegraph/cfg.cc#L501-L547
[...]

PR #1734, opened 2024-08-11
Support Windows... somewhat

wyattscarpenter:
In order to support windows, we need to use the c++-standard named struct params like @HexDecimal suggested in https://github.com/google/pytype/issues/1154#issuecomment-1108459343 and change cxx_std=11 to 20 like @rchen152 suggested in https://github.com/google/pytype/issues/1154#issuecomment-1377994508.

There may be more changes required. In particular, I am getting *one* test failing in the Tests step[^1] of the CI:
```
    File "D:\a/pytype/pytype/out\pytype\io_test.py", line 169, in test_unused_imports_info_files
      self.assertEqual(0, ret)
  AssertionError: 0 != 1
    ⋮
   FAILED: pytype/io_test.log D:/a/pytype/pytype/out/pytype/io_test.log
  C:\Windows\system32\cmd.exe /C "cd /D D:\a\pytype\pytype\out\pytype && C:\Python\Lib\site-packages\cmake\data\bin\cmake.exe -E env TYPESHED_HOME=D:/a/pytype/pytype/typeshed C:/Python/python.exe -B D:/a/pytype/pytype/build_scripts/test_module.py pytype.io_test -o D:/a/pytype/pytype/out/pytype/io_test.log -P D:/a/pytype/pytype/out -s -p"
  C:\Users\runneradmin\AppData\Local\Temp\tmp2coltc1x\m.py:1:1: error: in <module>: Can't find module 'common.foo'. [import-error]
  
  from common import foo; print(foo.x)~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  ```
 I don't know why. But this is still an incremental improvement in Windows support for pytype.
 
 [^1]: Additionally, in my [repo](https://github.com/wyattscarpenter/pytype/actions/runs/10342876412/job/28626335824#step:6:289), I've changed the CI script to run all of the checks and not early-out; in this, the Windows version now passes the Extensions Tests and gets some error in Type Check.

### CODE REVIEW

File: requirements.txt [hunk 1]
```diff
@@ -2,6 +2,7 @@
 # Make sure you also install the non-Python dependencies described in
 # https://github.com/google/pytype/blob/master/CONTRIBUTING.md#pytype-dependencies.
 attrs>=21.4.0
+cmake>=3.16
```
oprypin, line 5: I think we shouldn't be adding cmake here.

1. As I understand, it's a build-only dependency, whereas this file lists run-time dependencies. It is not necessary to install cmake to use pytype (I *think*)

2. It is rather controversial to install cmake from PyPI, or at least I haven't heard of this before.

3. For GitHub Actions it shouldn't be a problem, cmake is always available there

HexDecimal: I agree that `cmake` should at least be moved to `build-system.requires` in `pyproject.toml`, which already includes tools such as `ninja`.

I'd consider `cmake` and `ninja` to be equally controversial, being build tools external to Python's ecosystem, and `ninja` is already being installed from PyPI here.

oprypin: I see, indeed they are about equal. Sure.

Oh and also this file *is* for dev dependencies, we are not adding it to install dependencies.

OK, this is fine.

HexDecimal: My mistake. If it's only a dev dependency than it should be mentioned in `CONTRIBUTING.md` instead, which CMake already is. I'd agree with removing it from the requirements.txt in this case.

The CMake version in `CONTRIBUTING.md` should be bumped to `3.16` since that's the current minimum required version.

oprypin (dismissed):

requirements.txt
wyattscarpenter: Luckily, I already bumped the CMake version in CONTRIBUTING.md to 3.16! :)

I don't 100% understand which of these dependencies are used when, but moving cmake to build-system.requires works (moving ninja out of requirements.txt in the same way doesn't also work), so I have done so. Uh, by "works" I mean the CI fails on the same further step, of course.

HexDecimal: `requirements.txt`: suggested dependencies for package developers, in this case the ones working on pytype directly, would be installed into your local venv of your cloned project, sometimes automatically installed by your IDE.

`build-system.requires`: dependencies required to run `setup.py` or whichever `build-system.build-backend` is specified (currently setup.py), but are not needed to import the package normally. Do not move a dependency into `build-system.requires` unless the Python build step fails because that dependency is missing.

I haven't touched this project since #1154, having multiple suggested ways to build the project is a little confusing. `pip install -e .` would use `build-system.requires`, but the non-standard `python build_scripts/build.py` would expect the dependencies from `requirements.txt` instead. You are going to get different results depending on which build method is being used. I'm too much of an outsider to give any suggestions or advice on this, and my earlier advice may have been premature.

oprypin: I'm sorry, I think I'm not ready to accept these changes to build requirements within this pull request.

The decision how exactly cmake gets pulled in does not directly affect Windows support, and the situation is not unique to Windows; the support can be achieved without changing this. But it *does* affect every system in ways that we haven't fully understood yet.

So in the interest of getting this merged more easily, I am going to revert those changes from this pull request.

oprypin (approved):

oprypin:
Merged as https://github.com/google/pytype/commit/53135b9a15c3bdf1f452eae28c5875746ccbed52. Thank you

requirements.txt
wyattscarpenter: Sounds good. Thanks!

PR #1768, opened 2024-09-16
use jobs with pylint, speeding it up

wyattscarpenter:
This cuts down the time it takes to run lint by about a minute. "Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the number of processors available to use, and will cap the count on Windows to avoid hangs."


wyattscarpenter:
Good to hear!

oprypin:
And this was merged as https://github.com/google/pytype/commit/61363983e83e3fbfc23e9bda4b4e23a3d9badfdb

